PR #1, opened 2016-08-17
Bug 1296011 - Prototype for overriding user agents - DO NOT MERGE

denschub:
DO NOT MERGE.

This PR is a base for technological discussions in [bug 1296011](https://bugzilla.mozilla.org/show_bug.cgi?id=1296011).


bootstrap.js
```diff
@@ -0,0 +1,46 @@
...
+    let channel = subject.QueryInterface(Ci.nsIHttpChannel),
+      uaOverride = this.getUAForURI(channel.URI);
+
+    if (uaOverride) {
+      channel.setRequestHeader("User-Agent", uaOverride, false);
+    }
+  }
+
+  getUAForURI() {
+    return "GoFaster Test - " + DefaultUA;
```
MDTsai: I guess we will have a mapping table contains "URI":"UA"?

bootstrap.js
```diff
@@ -0,0 +1,46 @@
+/* global Components */
+
+const { classes: Cc, interfaces: Ci } = Components;
+const DefaultUA = Cc["@mozilla.org/network/protocol;1?name=http"]
+  .getService(Ci.nsIHttpProtocolHandler).userAgent;
+const ObserverService = Cc["@mozilla.org/observer-service;1"].getService(Ci.nsIObserverService);
+
+class UAOverrider {
+  init() {
+    ObserverService.addObserver(this, "http-on-useragent-request", false);
```
MDTsai: Suggest to use NS_HTTP_ON_USERAGENT_REQUEST_TOPIC instead of "http-on-useragent-request" directly?

denschub: I don't think we're able to use preprocessor constants in our JS sources, but feel free to enlighten me.

However, you raise a good point. I've moved the string to a const so we only have to change a single place if the topic ever changes.

bootstrap.js
denschub: Not too sure yet. Having a large list of URIs is one option, but if we do want to support globbing, we'd have to be a little bit more flexible.

At the moment, I'm thinking about a per-override validation function that receives the URL and can return a boolean which can trigger further processing. This may be a bit more complicated in the beginning, but having a URI->UA mapping table would be only an interim solution since we do have other requirements (like patching CSS or JS) as well. I'll investigate some ways to keep the performance impact low and start another discussion on Bugzilla when I'm ready.

miketaylr:
I've got a question about using jpm to build here... when we're shipping updates, this will be packaged up in moz-central via our `moz.build` file (https://github.com/mozilla/gecko-dev/blob/master/browser/extensions/webcompat/moz.build).

But maybe using `jpm` for dev is desirable? Not sure right now!

denschub:
Me neither, to be honest. `jpm` builds the `install.rdf` for me, but that's the only thing I am using at this moment. I don't use the jpm-built `bootstrap.js` since I need my own stuff in there.

The only reason to use `jpm` is that `jpm run` makes it fast and easy to test. We could put the `browser/extensions/webcompat` folder into an own git repository and symlink that external repository over `browser/extensions/webcompat`. `mach` has no issues with that and for some magic reasons, you don't even have to rebuild when running `./mach run`, so that could work as well. The only downside is that you need to have a m-c build on your machine and you have to update/rebuild it regularly. With jpm, you could simply use the nightlies from the build server.

denschub:
Oh, and hg would constantly complain about a dirty tree since all those files in `browser/extensions/webcompat` are removed as far as hg concerns.

miketaylr:
> We could put the browser/extensions/webcompat folder into an own git repository and symlink that external repository over browser/extensions/webcompat. mach has no issues with that and for some magic reasons, you don't even have to rebuild when running ./mach run, so that could work as well. The only downside is that you need to have a m-c build on your machine and you have to update/rebuild it regularly.

Yeah, the original intention was for this repo to be that external repository.

Having m-c on your machine and regularly building is a good thing, IMO. But maybe that's the stockholm symdrome speaking. 💀 😄 

I think a compromise would be possible, maybe a make file (or an npm script or whatever, I don't care) so you could do either. We just have to be sure to ignore that file (and the jpm stuff) when exporting diffs for syncing with m-c. Shouldn't be too hard -- and probably a makefile/whatever file task in itself.

denschub:
I have no issues with having m-c locally and updating it daily, but some contributors may. Who knows who will help us with site patches in the future. :)

++ for having a little script (I'd go with jake, just because makefiles scare me, if nobody objects) that generates a "clean" tree for diffing with the m-c copy. There are some files we'd have to ignore anyway, like the `.git` folder, readmes and eslint configs, which probably should not be in m-c. I'll build something before working on the per-URI override stuff!

miketaylr:
> I have no issues with having m-c locally and updating it daily, but some contributors may. 

Yeah, I get that. 

> ++ for having a little script (I'd go with jake, just because makefiles scare me, if nobody objects) that generates a "clean" tree for diffing with the m-c copy.

No objections from me. npm scripts are also an option, but if you're writing the patch you pick. :)

PR #2, opened 2016-09-06
Bug 1299448 - Create tools for exporting and running sources

denschub:
See Bugzilla.

r? @MDTsai


package.json
```diff
@@ -0,0 +1,22 @@
...
+    "jake": "./node_modules/.bin/jake"
+  },
+
+  "devDependencies": {
+    "eslint": "^3.4.0",
+    "eslint-plugin-json": "^1.2.0",
+    "eslint-plugin-mocha": "^4.5.1",
+    "eslint-plugin-mozilla": "0.0.3",
+    "eslint-plugin-promise": "^2.0.1",
+    "jake": "^8.0.12"
```
MDTsai: If we need node.js version > 5, I think we can add  "check-node-version" to devDependencies and run "check-node-version --node '>= 5.0.0'" before jake export-mc. Or we can make the node check in jake. Both are good for me.

src/bootstrap.js
```diff
@@ -4,7 +4,7 @@
  * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
 
 // for now we ship a stub, that can be upgraded as-needed after release
-function startup() {}
-function shutdown() {}
-function install() {}
-function uninstall() {}
+function startup() {} // eslint-disable-line no-unused-vars
```
MDTsai: I prefer not leave eslint comments to production code. Or update the function with parameters like:
```
function startup(data, reason) {
  // Do nothing in this version
}
```

package.json
denschub: I've just pushed a little change that ensures compatibility with Node 4, wich is the current LTS version. Support for v0.12 drops out in December, so 4.x is the oldest supported release. I guess we should be fine? :)

src/bootstrap.js
denschub: > I prefer not leave eslint comments to production code.

Me neither. However, `no-unused-vars` is not about the function being empty, it's about the function never getting called. `bootstrap.js` is not a Node module, so we cannot use `exported` here, so adding the comment is our only way to make eslint happy without turning `no-unused-vars` off for the entire file.

PR #3, opened 2016-09-28
Bug 1296011 - Override UAs on a per-URI basis

denschub:
See https://bugzilla.mozilla.org/show_bug.cgi?id=1296011.


src/content/lib/ua_overrider.js
```diff
@@ -0,0 +1,83 @@
...
+      if (!this._overrides[override.baseDomain]) {
+        this._overrides[override.baseDomain] = [];
+      }
+
+      if (!override.uriMatcher) {
+        override.uriMatcher = () => true;
+      }
+
+      this._overrides[override.baseDomain].push(override);
+    }
```
MDTsai: I prefer save override object instead push to an array object:
```javascript
    for (let override of overrides) {
      if (!override.uriMatcher) {
        override.uriMatcher = () => true;
      }

      this._overrides[override.baseDomain] = override;
    }
```

src/content/lib/ua_overrider.js
```diff
@@ -0,0 +1,83 @@
...
+
+  lookupUAOverride(uri) {
+    let baseDomain = eTLDService.getBaseDomain(uri);
+    if (this._overrides[baseDomain]) {
+      for (let uaOverride of this._overrides[baseDomain]) {
+        if (uaOverride.uriMatcher(uri.specIgnoringRef)) {
+          return uaOverride.uaTransformer(DefaultUA);
+        }
+      }
+    }
```
MDTsai: Then we can use the object directly here:
```javascript
if (this._overrides[baseDomain]) {
      let uaOverride = this._overrides[baseDomain];
      if (uaOverride.uriMatcher(uri.specIgnoringRef)) {
          return uaOverride.uaTransformer(DefaultUA);
      }
}
```
Not sure the difference between using a for loop to iterate uaOverride? Any performane issue?

MDTsai: For my test case (baseDomain: 163.com, test URI: http://sports.163.com), I can't intercept all network connects, like http://goal.sports.163.com/teamlogo/7495.png .

src/content/lib/ua_overrider.js
denschub: I expect there to be more than one override rule per base domain. Imagine overrides for Google Mail and Google maps, both are hosted on `google.com`, but with different `uriMatcher`s. That's why I use an array here.

src/content/lib/ua_overrider.js
denschub: As mentioned on IRC and [1296011#c1](https://bugzilla.mozilla.org/show_bug.cgi?id=1296011#c1), this is a known limitation of the way UA overrides get cached in Gecko. The override will only be determined and applied to the document URI request, not to any subsequent requests. This is how it works in Fennec at the moment, too.

I do not expect this to be an issue in the moment. If we ever run into a situation where we need to override the UA for a subresource, like an XHR or even a png file, we can discuss about how we get that done. I don't think this is a blocker for now. I also remember @miketaylr agreed with me here, but maybe we can get a second opinion? :)

miketaylr: Yeah, let's not block on that, but keep it in mind in case it because important in the future.

MDTsai: As it works fine here, @denschub could you add comment before function lookupUAOverride to describe the rule. That will be more clear for developers in the future.

denschub:
There is [ongoing discussion](https://mail.mozilla.org/pipermail/gofaster/2016-October/000448.html) about the future of Go Faster addons in regards of their underlying technology on the Go Faster mailing list. Basically, it says that Web Extensions should be used when possible (clearly not possible in our case) and that the Add-on SDK is deprecated and should be avoided.

The only call to the add-on SDK here is the `require("chrome")` inside `lib/ua_overrider.js`. I've done that since I use `require` in `bootstrap.js` to load the modules when needed. `Components.utils.import` could be used here, but it would require some workarounding. I guess we're fine for now, the `require` call could be replaced fairly easy if needed.

miketaylr:
> I've done that since I use require in bootstrap.js to load the modules when needed.

Is that something we could use `XPCOMUtils.defineLazyGetter` for?

miketaylr:
(actually, that should have been `XPCOMUtils.defineLazyModuleGetter`)

denschub:
There seem to be some issues with `defineLazyModuleGetter` and es6 classes, but I did not look into that deeper since I was fine with `require()`. Should I dig into that again?

miketaylr:
> There seem to be some issues with defineLazyModuleGetter and es6 classes, but I did not look into that deeper since I was fine with require(). Should I dig into that again?

Eh, maybe just file a follow up bug to investigate down the road. If the SDK is OK today, and possibly not OK down the road that's work we'll eventually have to do.

denschub:
Actually, nvm that. I'll look into that. Filing a follow-up seems a bit too much effort for a possibly small change.

denschub:
@MDTsai as discussed on IRC, I've made the module loading changes here. Tests are green, but it would be nice if you can have another look.

denschub:
Thank you very much, @MDTsai.

PR #4, opened 2016-10-21
Bug 1307792 - Add preference for disabling UA overrides

denschub:
r? @MDTsai

### DIFF
--- a/src/bootstrap.js
+++ b/src/bootstrap.js
@@ -2,22 +2,47 @@
  * License, v. 2.0. If a copy of the MPL was not distributed with this
  * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
 
-/* globals XPCOMUtils, UAOverrider UAOverrides */
+/* globals XPCOMUtils, Services, UAOverrider UAOverrides */
 
 Components.utils.import("resource://gre/modules/XPCOMUtils.jsm");
+
+const UA_ENABLE_PREF_NAME = "extensions.webcompat.perform_ua_overrides";
+
+XPCOMUtils.defineLazyModuleGetter(this, "Services", "resource://gre/modules/Services.jsm");
 XPCOMUtils.defineLazyModuleGetter(this, "UAOverrider", "chrome://webcompat/content/lib/ua_overrider.js");
 XPCOMUtils.defineLazyModuleGetter(this, "UAOverrides", "chrome://webcompat/content/data/ua_overrides.js");
 
 let overrider;
 
+function UAEnablePrefObserver() {
+  let isEnabled = Services.prefs.getBoolPref(UA_ENABLE_PREF_NAME);
+  if (isEnabled && !overrider) {
+    overrider = new UAOverrider(UAOverrides);
+    overrider.init();
+  } else if (!isEnabled && overrider) {
+    overrider.uninit();
+    overrider = false;
+  }
+}
+
 function install() {} // eslint-disable-line no-unused-vars
 function uninstall() {} // eslint-disable-line no-unused-vars
 
 function startup() { // eslint-disable-line no-unused-vars
+  // Intentionally set the preference to true on every browser restart to
+  // avoid site breakage by accidentally toggled preferences or by leaving
+  // it off after debugging a site.
+  Services.prefs.setBoolPref(UA_ENABLE_PREF_NAME, true);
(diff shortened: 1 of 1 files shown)


src/bootstrap.js
```diff
@@ -2,9 +2,13 @@
  * License, v. 2.0. If a copy of the MPL was not distributed with this
  * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
 
-/* globals XPCOMUtils, UAOverrider UAOverrides */
+/* globals XPCOMUtils, Services, UAOverrider UAOverrides */
 
 Components.utils.import("resource://gre/modules/XPCOMUtils.jsm");
+
+const UA_ENABLE_PREF_NAME = "extensions.webcompat.perform_ua_overrides";
```
miketaylr: (drive-by comment) 🚗 

So currently we only do UA overrides, so this seems fine. But in the future, do we envision a single pref to enable everything that this add-on can do? If so, does it make sense to make a more general pref?

(also, is the extensions namespace (prefix??) the right place for this? no idea!)

denschub: > If so, does it make sense to make a more general pref?

Maybe. Maybe it's best to have a pref per patch type. For now, this feels more natural at the moment. We can rename prefs if we'd like, so...

> is the extensions namespace (prefix??) the right place for this

That's what other Go-Faster extensions are doing (well, except for the e10srollout, they created their own prefix/namespace/whatever (branch, I guess), but as for pdfjs and hello, they sticked to `extensions.*`)

miketaylr: SGTM!

src/bootstrap.js
```diff
@@ -14,10 +18,25 @@ function install() {} // eslint-disable-line no-unused-vars
 function uninstall() {} // eslint-disable-line no-unused-vars
 
 function startup() { // eslint-disable-line no-unused-vars
+  Services.prefs.setBoolPref(UA_ENABLE_PREF_NAME, true);
```
MDTsai: I prefer not to enable preference in code, even if it's just in development stage. We should check preference then init UAOverrider, also add observer for changes.

src/bootstrap.js
```diff
@@ -14,10 +18,25 @@ function install() {} // eslint-disable-line no-unused-vars
...
+
   overrider = new UAOverrider(UAOverrides);
   overrider.init();
 }
 
 function shutdown() { // eslint-disable-line no-unused-vars
-  overrider.uninit();
+  if (overrider) {
+    overrider.uninit();
+  }
```
MDTsai: Another concern when shutdown, you can't use Services.prefs.removeObserver when it's an anonymous function. I suggest move observer to an independent function for removeObserver.

src/bootstrap.js
MDTsai: Maybe add a general pref in v2

src/bootstrap.js
denschub: We've discussed this in Taipei. :) The preference should reset on every browser reboot. If someone complains, we can change this in the future, but for now, we want to have the pref forced on on a browser startup.

MDTsai: If code forces on in m-c, then all users will be affected even in release channel. Could we consider force on in nightly but off in other versions? @miketaylr ? any comment?

MDTsai: Experience from other features, when in developing stage, it's better not to directly turned on. Developers can easily turned on by 1) setup about:config or 2) patch preference file.
When code is checked-in to m-c, and it's not completed or verified by QA, at least we can promise it will not affect other features.
After implementation completed and ready for QA, we can submit a patch to user.js to turn on.
If we still prefer directly turn on in code, please add a nightly or developer edition detection check to prevent it will not affect beta and official release.
@denschub It will prevent in-complete implementation checked-in to m-c and affect something we don't want. I agree to turn on this feature in Dev Edition and Nightly, but please do this after implementation complete and approved by release manager. Does this make sense?

src/bootstrap.js
denschub: Great, thanks for that. I've changed that. :)

miketaylr:
What's the status on this PR @denschub @MDTsai?

src/bootstrap.js
MDTsai: @denschub  a nit, please add comment here to let other developers know we force to reset this preference on purpose.

denschub: Right! Added a comment. :)

PR #5, opened 2016-11-01
Bug 1307788 - Implement Developer Tools notification

MDTsai:
Embed a WebExtension to receive tab update event.
While tab is loading and the URL matches override rule,
Injects a javascript to print console log.
Uses chrome.runtime.onMessage to link WebExtension and UAOverrider.

### DIFF
--- a/src/bootstrap.js
+++ b/src/bootstrap.js
@@ -13,6 +13,7 @@ XPCOMUtils.defineLazyModuleGetter(this, "UAOverrider", "chrome://webcompat/conte
 XPCOMUtils.defineLazyModuleGetter(this, "UAOverrides", "chrome://webcompat/content/data/ua_overrides.js");
 
 let overrider;
+let tabUpdateHandler;
 
 function UAEnablePrefObserver() {
   let isEnabled = Services.prefs.getBoolPref(UA_ENABLE_PREF_NAME);
@@ -28,7 +29,7 @@ function UAEnablePrefObserver() {
 function install() {} // eslint-disable-line no-unused-vars
 function uninstall() {} // eslint-disable-line no-unused-vars
 
-function startup() { // eslint-disable-line no-unused-vars
+function startup({webExtension}) {
   // Intentionally set the preference to true on every browser restart to
   // avoid site breakage by accidentally toggled preferences or by leaving
   // it off after debugging a site.
@@ -37,8 +38,33 @@ function startup() { // eslint-disable-line no-unused-vars
 
   overrider = new UAOverrider(UAOverrides);
   overrider.init();
+
+  // Init webExtension to listen tab update status
+  webExtension.startup().then((api) => {
+    const {browser} = api;
+    // tabUpdateHandler receives tab updated event from WebExtension tablog.js
+    // While tab status changes to loading, tablog.js queries this URI is overrided or not.
+    // tabUpdateHandler uses sendResponse sends result to tablog.js
+    // tablog.js can determine to print web console log or not.
+    tabUpdateHandler = function(message, sender, sendResponse) {
+      try {
+        if (overrider) {
+          let uaOverride = overrider.getUAForURI(Services.io.newURI(message.url, null, null));
+          sendResponse({reply: !!uaOverride});
+        }
+      } catch (exception) {
+        sendResponse({reply: false});
+      }
+    };
+
+    browser.runtime.onMessage.addListener(tabUpdateHandler);
+    return;
+  }).catch((reason) => {
+    console.log(reason);
+  });
 }
 
+// TODO: Figure out how to remove listener when bootstrapped addon shutdown
 function shutdown() { // eslint-disable-line no-unused-vars
   Services.prefs.removeObserver(UA_ENABLE_PREF_NAME, UAEnablePrefObserver);
 
--- a/src/webextension/manifest.json
+++ b/src/webextension/manifest.json
@@ -1,0 +1,16 @@
+{
+
+  "description": "WebCompat Go Faster Web Extension",
+  "manifest_version": 2,
+  "name": "webcompat-extension",
+  "version": "1.0",
+
+  "permissions": [
+    "tabs",
+    "<all_urls>"
+  ],
+
+  "background": {
+    "scripts": ["tablog.js"]
+  }
+}
--- a/src/webextension/tablog.js
+++ b/src/webextension/tablog.js
@@ -1,0 +1,23 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+"use strict";
+
+const consoleLogScript = 'console.log("The user agent has been overridden for compatibility reasons.");';
+
+/* This function handle tab updates.
+ * While changeInfo.status is loading and there is url
+ * Send tab url to ua_overrider.js to query if needs to print web console log.
+ */
+function handleUpdated(tabId, changeInfo, tabInfo) {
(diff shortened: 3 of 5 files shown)


miketaylr:
Would it make sense to keep a Map/Object/Set/Whatever to associate bug numbers to URIs we override? Then we use that to dump the right URL in the console.log message?

(It would be a manual thing we have to update, but maybe that's not terrible?)

MDTsai:
Yes, @denschub and I had a plan to implement this. So this is a temp patch.
This patch prove we can connect XPCOM from chrome process to WebExtension in content process.

src/webextension/manifest.json
```diff
@@ -0,0 +1,18 @@
...
+  "description": "WebCompat Go Faster Web Extension",
+  "manifest_version": 2,
+  "name": "webcompat-extension",
+  "version": "1.0",
+
+  "permissions": [
+    "tabs",
+    // TODO: Fill required URL for override rules the same as ua_overrides.js
+    // For tablog.js to have permission to inject script to web console.
+    // Example: "*://www.google.com.tw/*"
```
denschub: Do you think it's possible to automatically populate this list somehow? I am not too keen on having to maintain two lists of domains, especially since they don't share the same format.

src/bootstrap.js
```diff
@@ -2,22 +2,44 @@
...
 XPCOMUtils.defineLazyModuleGetter(this, "UAOverrider", "chrome://webcompat/content/lib/ua_overrider.js");
 XPCOMUtils.defineLazyModuleGetter(this, "UAOverrides", "chrome://webcompat/content/data/ua_overrides.js");
 
 let overrider;
+let tabUpdateHandler;
 
 function install() {} // eslint-disable-line no-unused-vars
 function uninstall() {} // eslint-disable-line no-unused-vars
 
-function startup() { // eslint-disable-line no-unused-vars
```
denschub: Nit: please keep the eslint comment, otherwise, eslint complains:

> 'startup' is defined but never used

src/bootstrap.js
```diff
@@ -2,22 +2,44 @@
...
+function startup({webExtension}) {
   overrider = new UAOverrider(UAOverrides);
   overrider.init();
+
+  // Init webExtension to listen tab update status
+  webExtension.startup().then(api => {
```
denschub: Can this ever throw? If so, we should probably add a `catch()` here. (also, that's what our eslint rules suggest)

denschub: Nit: our eslint rules enforce parentheses around arrow function arguments.

src/bootstrap.js
```diff
@@ -2,22 +2,44 @@
...
+    const {browser} = api;
+    // tabUpdateHandler receives tab updated event from WebExtension tablog.js
+    // While tab status changes to loading, tablog.js queries this URI is overrided or not.
+    // tabUpdateHandler uses sendResponse sends result to tablog.js
+    // tablog.js can determine to print web console log or not.
+    tabUpdateHandler = function (message, sender, sendResponse) {
```
denschub: > Unexpected space before function parentheses

src/bootstrap.js
```diff
@@ -2,22 +2,44 @@
...
+      try {
+        let uaOverride = overrider.getUAForURI(Services.io.newURI(message.url, null, null));
+        sendResponse({reply: (uaOverride ? true : false)});
```
denschub:
```js
sendResponse({reply: !!uaOverride});
```

src/webextension/tablog.js
```diff
@@ -0,0 +1,23 @@
...
+"use strict";
+
+const consoleLogScript = 'console.log("The user agent has been overridden for compatibility reasons.");';
+
+/* This function handle tab updates.
+ * While changeInfo.status is loading and there is url
+ * Send tab url to ua_overrider.js to query if needs to print web console log.
+ */
+function handleUpdated(tabId, changeInfo, tabInfo) {
+  if (changeInfo.status == "loading" && changeInfo.url) {
```
denschub: Nit: `===` instead of `==`

src/webextension/tablog.js
```diff
@@ -0,0 +1,23 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+"use strict";
```
denschub: Should probably add a `globals` definition for eslint here, since it complains that `chrome` is undefined.

src/webextension/tablog.js
```diff
@@ -0,0 +1,23 @@
...
+
+/* This function handle tab updates.
+ * While changeInfo.status is loading and there is url
+ * Send tab url to ua_overrider.js to query if needs to print web console log.
+ */
+function handleUpdated(tabId, changeInfo, tabInfo) {
+  if (changeInfo.status == "loading" && changeInfo.url) {
+    chrome.runtime.sendMessage( {url: changeInfo.url}, (message) => {
+      if (message.reply) {
+        chrome.tabs.executeScript(tabId, { code: consoleLogScript });
```
denschub: > 17:42  error  There should be no space after '{'           object-curly-spacing
> 17:67  error  There should be no space before '}'          object-curly-spacing

src/webextension/tablog.js
```diff
@@ -0,0 +1,23 @@
...
+
+const consoleLogScript = 'console.log("The user agent has been overridden for compatibility reasons.");';
+
+/* This function handle tab updates.
+ * While changeInfo.status is loading and there is url
+ * Send tab url to ua_overrider.js to query if needs to print web console log.
+ */
+function handleUpdated(tabId, changeInfo, tabInfo) {
+  if (changeInfo.status == "loading" && changeInfo.url) {
+    chrome.runtime.sendMessage( {url: changeInfo.url}, (message) => {
```
denschub: > There should be no spaces inside this paren

src/webextension/manifest.json
MDTsai: It's possible to use "<all_urls>", I think you prefer this way

denschub: That would mean the script would be injected in all websites, no? Does this have any performance implications?

What I was initially thinking about was a way of setting the list of trusted URLs after we loaded the list of overwrites. Is there any way we can set this during runtime?

MDTsai: @denschub: yes, all_urls is a special host permission, we can refer to:
https://developer.mozilla.org/en-US/Add-ons/WebExtensions/manifest.json/permissions#Host_permissions
https://developer.mozilla.org/en-US/Add-ons/WebExtensions/Match_patterns#%3Call_urls%3E

I think the permission check is on-air, while I want to inject script, ExtensionContent.jsm will check and through exception. All WebExtension use this design, so far I don't see performance issue here.

For manifest.json, it can't be set on runtime. But we can set from ua_overrides.js before we install to gecko-dev by jake.

denschub: > so far I don't see performance issue here.

Okay, cool, thanks. :)

> But we can set from ua_overrides.js before we install to gecko-dev by jake.

That sounds good. Do you want to implement that as well? Feels much nicer than having to maintain two lists of domain.

src/webextension/tablog.js
MDTsai: fixed

src/webextension/tablog.js
MDTsai: fixed

src/webextension/tablog.js
MDTsai: hmm, I guess "chrome" is already defined as global while init WebExtension. Like startup is not defined in bootstrap.js

denschub:
Thanks. 🎉

PR #6, opened 2016-11-09
No Bug - Fix up some docs

vladikoff:
Found some issues with docs while investigating `Go Faster` add-ons.. 👀


miketaylr:
Thanks @vladikoff!

PR #8, opened 2016-11-17
WebCompat 3714 - Add override for Google Play Newsstand

denschub:
See https://github.com/webcompat/web-bugs/issues/3714

Added a comment block for explanation as long as we don't have a nice way of storing the webcompat bug ID etc. in there. As per our policy draft, I've tested some different user agents and appending the Chrome fragments to the existing Firefox fragments seems to trick their UA detection.

r? @MDTsai 
feedback? @miketaylr

### DIFF
--- a/src/content/data/ua_overrides.js
+++ b/src/content/data/ua_overrides.js
@@ -39,6 +39,18 @@
  */
 
 const UAOverrides = [ // eslint-disable-line
+  /*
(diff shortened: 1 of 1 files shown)


denschub:
Eh, cool. This just got obsoleted by Google being faster than us. Great. 😿 

Will look for a new test case tomorrow morning.

PR #10, opened 2017-01-05
Bug 1328853 - Don't always print the console message

denschub:
See [bug 1328853](https://bugzilla.mozilla.org/show_bug.cgi?id=1328853).


denschub:
r+ from @MDTsai on bugzilla, merged.

PR #11, opened 2017-01-05
Web bug 4032 - Override UA for Netflix on Linux

denschub:
See https://github.com/webcompat/web-bugs/issues/4032.

Netflix uses a server-side feature detection on their watch page. On Firefox Linux, they claim we don't support HTML5 video with DRM, although the support for Widevine on Firefox Linux is good enough for Netflix these days.

Their check only looks for a Chrome snippet in the user agent, so we're fine with simply appending it.

Note that this will only return a modified UA on Linux installations, since other operating systems are already officially supported by Netflix.

r? @miketaylr for both code review and a verification of this UA override.

### DIFF
--- a/src/content/data/ua_overrides.jsm
+++ b/src/content/data/ua_overrides.jsm
@@ -54,6 +54,33 @@ const UAOverrides = [
       let prefix = originalUA.substr(0, originalUA.indexOf(")") + 1);
       return `${prefix} AppleWebKit/537.36 (KHTML, like Gecko) Chrome/54.0.2840.98 Safari/537.36`;
     }
+  },
+
+  /*
+   * web-bug #4032 - video doesn't play on netflix.com
+   * https://github.com/webcompat/web-bugs/issues/4032
+   *
+   * Netflix uses a server-side feature detection on their watch page. On
+   * Firefox Linux, they claim we don't support HTML5 video with DRM, although
+   * the support for Widevine on Firefox Linux is good enough for Netflix these
+   * days.
+   *
+   * Their check only looks for a Chrome snippet in the user agent, so we're
+   * fine with simply appending it.
+   *
+   * Note that this will only return a modified UA on Linux installations, since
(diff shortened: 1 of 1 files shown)


src/content/data/ua_overrides.jsm
```diff
@@ -54,6 +54,33 @@ const UAOverrides = [
...
+   * fine with simply appending it.
+   *
+   * Note that this will only return a modified UA on Linux installations, since
+   * other operating systems are already officially supported by Netflix.
+   */
+  {
+    baseDomain: "netflix.com",
+    uriMatcher: (uri) => uri.includes("/watch/"),
+    uaTransformer: (originalUA) => {
+      if(!originalUA.includes("Linux")) {
```
denschub: I added an easier platform-filter to version 1, see https://wiki.mozilla.org/Compatibility/Go_Faster_Addon/Roadmap#Version_2.

miketaylr:
OK, once I installed `libav-tool` this works as expected. 

@cpearce, curious what you think here. If the user can't handle the mp4 (is this normal, or just a VM thing?), would sending them to this Chrome error page be worse than the support page that tells them they're not supported?

Assuming we could figure out a `canPlayType` check for the codecs from the error message, what I wonder is that stable to ship? Or dependent on any number of variable (or prone to change)?

`var f = document.createElement("video"); f.canPlayType("video/mp4;codecs=avc1.640028")`

denschub:
Just for the record, the default behavior without the UA override is a redirect to https://help.netflix.com/en/node/23742.

cpearce:
I expect you were failing to play without libav-tool because you lacked an AAC audio codec, rather than an H.264/AVC codec. Netflix's audio streams are not encrypted, so Firefox requires an installed system audio codec to play Netflix. Most Linux distros don't install one by default.

miketaylr:
Let's go ahead and close this PR for now, possibly we'll come back to it.

PR #12, opened 2017-01-09
Fix mochitest

MDTsai:
Change use defineLazyModuleGetter from ua_overrider.js to ua_overrider.jsm


denschub:
Hmhm. I've merged this, but I just now realized we have no bug number in the commit message. Can we agree on having `Bug xxx - Foo bar` commit messages like we've started? :)

miketaylr:
+1 to using Gecko style commit messages.

PR #13, opened 2017-02-08
Bug 1337782 - Remove URI caching

denschub:
r? @MDTsai, see [bug 1337782](https://bugzilla.mozilla.org/show_bug.cgi?id=1337782)

### DIFF
--- a/src/content/lib/ua_overrider.jsm
+++ b/src/content/lib/ua_overrider.jsm
@@ -16,7 +16,6 @@ XPCOMUtils.defineLazyServiceGetter(this, "eTLDService", "@mozilla.org/network/ef
 class UAOverrider {
   constructor(overrides) {
     this._overrides = {};
-    this._overrideForURICache = new Map();
 
     this.initOverrides(overrides);
   }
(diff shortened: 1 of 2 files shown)


src/content/lib/ua_overrider.jsm
```diff
@@ -59,34 +58,7 @@ class UAOverrider {
 
   getUAForURI(uri) {
     let bareUri = uri.specIgnoringRef;
```
MDTsai: bareUri is not used, please remove unused variable.
As bareUri is removed, I think we don't need getUAForURI just for calling lookupUAOverride.

denschub:
You are very right. Fixed. Also, rebased on latest master.

MDTsai:
@denschub, I fail to run the unit test. Please also fix the test case then I think we are ready to merge the patch.

denschub:
Not sure if you saw the IRC message, but this is fixed. :)

denschub:
Oh, nvm, you r+'d on bugzilla. Should've read my emails first. Thanks for the review!

PR #14, opened 2017-02-08
Bug 1337785 - Don't act on URIs where we can't get the baseDomain

denschub:
r? @MDTsai, see [bug 1337785](https://bugzilla.mozilla.org/show_bug.cgi?id=1337785)


denschub:
Comment added, thanks for the review.

PR #16, opened 2017-04-26
Bug 1316668 - Re-add devtools looking and make it better

denschub:
Sorry that this contains two changes in one PR - I already spent quite some time with splitting off other parts, and there was no way I can sensibly not PR these two at the same time (I'd go into rebasing hell for that one, so...)

---

After a long and hard journey (of which I should write a book about), here is a PR that looks like 10 minutes of work. :) Basically, this uses `nsIConsoleService` to send a custom message to the console. Also, I added the option of adding a link to the contents, since we need that one anyway.

I may change the format again later, I am currently playing around with a possible way to make the link actually clickable. But if this gets reviewed by @MDTsai before I finished that, I'll follow up. For now, I just want to have this merged.

r? @MDTsai

### DIFF
--- a/src/content/data/ua_overrides.jsm
+++ b/src/content/data/ua_overrides.jsm
@@ -8,6 +8,9 @@
  *
  *   * `baseDomain`, required: The base domain that further checks and user
  *     agents override are applied to. This does not include subdomains.
+ *   * `moreInfo`: An URL to a web page that contains more information about
+ *     why the override was put in place and explaining how the override can
+ *     be disabled.
  *   * `uriMatcher`: Function that gets the requested URI passed in the first
  *     argument and needs to return boolean whether or not the override should
  *     be applied. If not provided, the user agent override will be applied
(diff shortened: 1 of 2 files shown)


miketaylr:
Ping @MDTsai

MDTsai:
@denschub I think you need a update for this? 
```
DenSchub> etsai: hold back with the review for one more day -- i want to push another thing. i thought i could finish it tonight, but it's 1am now and i'm too tired. i'll ping you tomorrow. :)
```

miketaylr:
(oops, sorry looks like i was butting in :P)

miketaylr:
> hold back with the review for one more day -- i want to push another thing.

@denschub did you forget to push this thing?

denschub:
> did you forget to push this thing?

No, I don't. it ended up regressing perf tests, then I ended up fixing quantum startup, then other work happened, then I closed this PR by accident. Reopening and noting that this is on top of my stack after landing the fixes.

denschub:
Deprecated, a new PR will be opened soon.

PR #19, opened 2017-07-25
Bug 1383989 - Add a plattform filter for overrides

denschub:
r? @MDTsai - please see [bugzilla](https://bugzilla.mozilla.org/show_bug.cgi?id=1383989) for more information.

### DIFF
--- a/Jakefile
+++ b/Jakefile
@@ -161,17 +161,30 @@ function generateChromeManifest() {
   });
 }
 
+/**
+ * Exports the files to a target, used to export into mozilla-central
+ *
+ */
+function exportFiles(target) {
+  let mcLocation = getMozillaCentralLocation();
+  let extTargetDir = path.join(mcLocation, target);
+  jake.rmRf(extTargetDir);
+  jake.cpR(BUILD_DIR, extTargetDir);
+
+  console.log(`Exported built sources into ${extTargetDir}`);
+}
+
 desc(`Builds the extension into the ${BUILD_DIR}/ directory`);
 task("build", ["building:cleanup", "building:copy"], () => {});
 
 desc("Exports the sources into mozilla-central");
 task("export-mc", ["build"], () => {
-  let mcLocation = getMozillaCentralLocation();
-  let extTargetDir = path.join(mcLocation, "browser/extensions/webcompat");
-  jake.rmRf(extTargetDir);
-  jake.cpR(BUILD_DIR, extTargetDir);
+  exportFiles("browser/extensions/webcompat");
+});
 
-  console.log(`Exported built sources into ${extTargetDir}`);
+desc("Exports the sources into the mozilla-central for android");
+task("export-mc-android", ["build"], () => {
+  exportFiles("mobile/android/extensions/webcompat");
 });
 
 desc("Exports the sources into an .xpi for update shipping");
--- a/src/content/lib/ua_overrider.jsm
+++ b/src/content/lib/ua_overrider.jsm
@@ -19,7 +19,27 @@ class UAOverrider {
   }
 
   initOverrides(overrides) {
+    // on xpcshell tests, there is no impleentation for nsIXULAppInfo, so this
+    // might fail there. To have all of our test cases running at all times,
+    // assume they are on Desktop for now.
+    let currentApplication = "firefox";
+    try {
+      let appInfo = Cc["@mozilla.org/xre/app-info;1"].getService(Ci.nsIXULAppInfo);
+      currentApplication = appInfo.name.toLowerCase();
+    } catch(_) {}
+
     for (let override of overrides) {
+      // Firefox for Desktop is the default application for all overrides.
+      if (!override.applications) {
+        override.applications = ["firefox"];
+      }
+
+      // If the current application is not targeted by the override in question,
+      // we can skip adding the override to our checks entirely.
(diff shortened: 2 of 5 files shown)


src/content/lib/ua_overrider.jsm
```diff
@@ -19,7 +19,27 @@ class UAOverrider {
...
     for (let override of overrides) {
+      // Firefox for Desktop is the default application for all overrides.
+      if (!override.applications) {
+        override.applications = ["firefox"];
+      }
+
+      // If the current application is not targeted by the override in question,
+      // we can skip adding the override to our checks entirely.
+      if (!override.applications.includes(currentApplication)) {
+        continue;
```
MDTsai: Should we consider "Request desktop site" for fennec here? @miketaylr

Jakefile
```diff
@@ -161,17 +161,30 @@ function generateChromeManifest() {
...
-  let mcLocation = getMozillaCentralLocation();
-  let extTargetDir = path.join(mcLocation, "browser/extensions/webcompat");
-  jake.rmRf(extTargetDir);
-  jake.cpR(BUILD_DIR, extTargetDir);
+  exportFiles("browser/extensions/webcompat");
+});
 
-  console.log(`Exported built sources into ${extTargetDir}`);
+desc("Exports the sources into the mozilla-central for android");
+task("export-mc-android", ["build"], () => {
```
MDTsai: We need to support fennec on android, but with 2 export-mc-* functions means we have 2 source codes in mc?
First idea 2 platform shares 1 code, so we do only need 1 copy.
Second idea is like flyweb, they maintain 2 codes, then we use application info to export to folder. 
Does webcompat reporter extension has the same problem for you @miketaylr ?

MDTsai:
Could we mock something to test both applications in XPCShell?

Jakefile
denschub: The "report site issue" thing on Fennec is not an extension, it [is a module natively built into Fennec](https://dxr.mozilla.org/mozilla-central/source/mobile/android/chrome/content/WebcompatReporter.js).

> First idea 2 platform shares 1 code, so we do only need 1 copy.

I would love to share the code between both builds, but system extensions for Desktop are in `browser/extensions` while system extensions for Fennec are in `mobile/android/extensions`. I had a quick look at the build chain, and there seems to be no link between those two builds in that regards. How would you share the sources?

> Second idea is like flyweb, they maintain 2 codes, then we use application info to export to folder.

I don't think I understand what you mean here. We export the exact same sources for both desktop and mobile, there is no actual code difference. The reason they can get exported individually is the fact that landing a system extension update in Desktop is much more complicated process-wise, and QA steps and thus landing timelines may differ. Could you elaborate what you mean here? :)

MDTsai: For sharing 1 copy, if there is no possible way, I think the simple way is to export source to 2 folder.
For 2 copies, that means we don't need to copy rule for "fennec" to desktop. We can use the application property in ua_overrides.jsm. May we can have a good way, like use macro/define to preprocess, then ua_overrider.jsm don't need to filter in initialization stage.

denschub: > For sharing 1 copy, if there is no possible way, I think the simple way is to export source to 2 folders.

Which is exactly what we are doing here, no?

denschub: Also, as an addition: updates for this extension can and will be served over addons.mozilla.org and the updated via the auto updater. Having different addon contents (like, for example, modifying `ua_overrides.jsm` in a preprocessor), would mean we'd actually end up having two independent extensions, `webcompat-desktop@mozilla.org` and `webcompat-android@mozilla.org`, which means that, for every update, we'd need two QA signoffs and duplicating a lot of other work. I don't see how that's better than copying the exact same source into two folders in mozilla-central.

I feel like I am missing your point here. Mind hopping on Vidyo tomorrow to have a quick discussion on that? :)

MDTsai: So we only have one extension for both desktop and android.

What I think is we only need to export-mc, which export file to **browser/extensions/webcompat'** and **mobile/android/extensions/webcompat** no matter what change we make, to prevent we don't miss any change to commit to m-c. But I don't know if other reviewer will have concern to put same source code in 2 different folders.

denschub: I am very sure that people who are going to review our updates to the desktop tree are not willing/able to review the mobile changes as well, so landing changes for fennec and desktop has to happen in two bugs, with two commits and two sets of reviewers. That's the reason I wrote a second exporter.

> But I don't know if other reviewer will have concern to put same source code in 2 different folders.

I talked to some people and so far, it seems like nobody has an issue with that since there is no other possibility to share sources. however, we are the first people to ever do that, so it may be that there will be additional discussion on landing. If it turns out we need to do something else, we can do that, but for now, this seems good enough.

src/content/lib/ua_overrider.jsm
miketaylr: Do you mean not apply the override when that's enabled? I would assume so.

Jakefile
MDTsai: Input from others if we want to put 1 source code in m-c:
1. Use gecko-dev/toolkit
2. Dev-tool team plans to move to system addon, they may have the problem as us.

denschub:
welp, I totally did not mean to close this.

denschub:
Deprecated, a new PR will be opened soon.

PR #20, opened 2018-02-02
Combined PR for 1383989/1338586/1404463 (GoFaster v2)

denschub:
This one is for @wisniewskit, who is not yet part of the GitHub org, so I'll have to stick with a written `r? @wisniewskit`. :)

Sorry this is all in one PR, but since I was changing earlier commits regularly, it did not made sense to me to submit individual PRs, since they all depend on each other. The commits are well separated, so maybe reviewing each commit is a good idea. If you need help or prefer having separate PRs, just let me know, I can do that.

There is a [wiki page](https://wiki.mozilla.org/Compatibility/Go_Faster_Addon/Override_Policies_and_Workflows) going into detail about the "config" files and other technical details. Might be worth checking out as well.

Some words to each part, and to provide more context.

* The `Add a platform filter for overrides` and the following build tools commit were previously PR'ed, but not merged. Eric noted that it may be bad to duplicate the addon sources for Desktop and Mobile. I had a chat with some people, and although I did not get an "official" answer, everyone basically told me they want to keep the amount of sources shared between both products as low as possible, so duplicating may actually be preferable. Together with Mike, I came up with the idea that we will just *try* shipping it, and do adjustments if it turns out that duplicating sources is an issue.
* For 1404463, I had an entirely different approach at first, but it turned out to be hacky, sometimes racy, and not very efficient. So I decided to 🗑 that and just roll with webextensions. Ultimately, this gives a nicer and simpler solution of injecting CSS and JS anyway, but for the downside that we now have two places to configure stuff in: UA overrides in the legacy extension, CSS/JS injections in the WebExtension. However, since we could do our UA overrides in the WebExtension as well these days, this is probably just an interim downside.
* The "Switch the UA override test over to the new test case site " commit is just maintenance. I built a new site for JS/CSS injection tests (mainly for QA), so I figured it would be good to just have one site to test everything.
* Also, I moved the docs out of the addon completely and into a MozWiki page. The reason is explained in the commit, but in short: we can't have comments in the WebExtensions manifest anyway, so we'd need a dedicated place for docs. Since having two places for documentation is bad, I just moved everything into a wiki page.

As for testing, you have two options, check the readme. If you run `npm run jake export-mc`/`npm run jake run-mc`, you can test the sources as they would be when landed inside Gecko. The downside, however, is that you have to have an uptodate m-c build, and rebuild whenever the sources change. `npm run jake export-xpi` will build you an .xpi inside the `build/` folder to install with `extensions.legacy.enabled` turned on and `xpinstall.signatures.required` turned off. However, since the other version is loaded as a system extension, it does not cleanly update it, but install in parallel. Because of that, you have to reboot Firefox after installing the extension to make the system extension's impact disappear... I'll check if there is a bug filed for this already, and file one if not, as it makes testing rather annoying. Just a heads-up. :)

---

Open ToDos

- [ ] Adopt [`ChromeUtils.import`](https://groups.google.com/forum/#!topic/mozilla.dev.platform/xQaTdsrbd-g). Although this will break the extension for earlier Firefox versions, if we do the adoption in a single commit, we could revert that if we have to build and ship an XPI out-of-band.


wisniewskit:
That's an r+ from me on these patches. My only suggestions are
- Use `window.eval` now instead of `window.wrappedJSObject.eval` (which seems to work fine for me in local testing).
- I'm not sure, but to cope with the recent lazy module importing changes, we may need to use code-blocks like this (working locally for me on old and new m-c builds):
```
if (XPCOMUtils.defineLazyModuleGetter) {
  XPCOMUtils.defineLazyModuleGetter(this, "UAOverrider", "chrome://webcompat/content/lib/ua_overrider.jsm");
  XPCOMUtils.defineLazyModuleGetter(this, "UAOverrides", "chrome://webcompat/content/data/ua_overrides.jsm");
} else {
  ChromeUtils.defineModuleGetter(this, "UAOverrider", "chrome://webcompat/content/lib/ua_overrider.jsm");
  ChromeUtils.defineModuleGetter(this, "UAOverrides", "chrome://webcompat/content/data/ua_overrides.jsm");
}
```
- Follow-ups to port as much of the addon into a pure WebExtension as possible (which I know is planned).

denschub:
@wisniewskit I fixed your remarks, would be nice if you could verify. :)

As for the pref to turn injections off (as discussed on IRC), I need your help here. While I have no issue sending messages to the background script, I didn't quite get your idea about removing the event handler. If I completely remove all handlers, I'd have to re-implement the default by hand, as far as I can see. I'm probably missing something simple here. Could you go into detail and explain how you would disable the content scripts inside the background script? Ignore the messaging part, I've figured that one out. Thanks.

denschub:
The pref issue was resolved via IRC. We're going to not define the contentScripts in the manifest, but manually registering them with `contentScripts.register`, which allows us to unregister them at a later point (i.e. after we get a message to disable them).

wisniewskit:
For the record, the other follow-up changes are fine; just one note I mentioned on IRC.

denschub:
Updated the PR to include the pref. @wisniewskit, r? again, hopefully for the last time. All changes between the last review and now were made in commit a02fc9cfdc7ff445a32984792debaeeb2389a9f3, the rest is untouched (besides a rebase).

Documentation is outdated right now, since I moved the location for defining content scripts, but I'll change this when this has r+ and Gecko review is on its way.

wisniewskit:
Looks good to me! Just bear in mind that contentScripts.register/unregister aren't actually available until Firefox 59.0b5, so you might want to tweak your minimum versions accordingly.

denschub:
Minimum version changed, and merged. Thank you very much for your reviews.

PR #21, opened 2018-04-10
Bug 1452707 - Add WebKit-like UA Override for ib.absa.co.za

denschub:
See [bug 1452707](https://bugzilla.mozilla.org/show_bug.cgi?id=1452707) for details.

r? @wisniewskit for the technical part,  
r? @miketaylr for the UA override, and as a second reviewer.

### DIFF
--- a/src/content/data/ua_overrides.jsm
+++ b/src/content/data/ua_overrides.jsm
@@ -54,6 +54,22 @@ const UAOverrides = [
       let prefix = originalUA.substr(0, originalUA.indexOf(")") + 1);
       return `${prefix} AppleWebKit/537.36 (KHTML, like Gecko) Chrome/54.0.2840.98 Safari/537.36`;
     }
+  },
+
+  /*
+   * WebCompat bug #16401
+   * https://webcompat.com/issues/16401
+   *
+   * Starting with Firefox 61, the window.controllers shim will no longer be
+   * available, causing ib.absa.co.za to show a "Browser unsupported" message
+   * and blocking the page.
+   */
+  {
+    baseDomain: "absa.co.za",
+    uriMatcher: (uri) => uri.includes("ib.absa.co.za"),
+    uaTransformer: (originalUA) => {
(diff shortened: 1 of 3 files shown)


src/content/data/ua_overrides.jsm
```diff
@@ -54,6 +54,22 @@ const UAOverrides = [
...
+   *
+   * Starting with Firefox 61, the window.controllers shim will no longer be
+   * available, causing ib.absa.co.za to show a "Browser unsupported" message
+   * and blocking the page.
+   */
+  {
+    baseDomain: "absa.co.za",
+    uriMatcher: (uri) => uri.includes("ib.absa.co.za"),
+    uaTransformer: (originalUA) => {
+      return `${originalUA} AppleWebKit/605.1.15 for Compatibility`;
```
miketaylr: The `for Compatibility` bit is interesting! I wonder if we shouldn't have it in general though, as it might confuse certain UA parsing libs? Thoughts @wisniewskit?

miketaylr: @denschub can you remind me the status of console logging a message for the site developers?

wisniewskit:
I'm honestly not sure if adding "for Compatibility" would break anything, but that's true of any UA fix in general, so it's probably worth making such changes site-specific. For the record, I've been adding "FxQuantum/v.v.v.v" to the UA string in my Google Search Fixer on Google (it's otherwise the same as Chrome's UA).

src/content/data/ua_overrides.jsm
denschub: We currently end up in the browser console. We have bug 1316668, but that didn't make it into v2 in our triage some months ago.

denschub:
Just for the record, if v2 lands, we could also just shim window.controllers back into existence. Any non-falsey value would work for the site to work again, and we would not have to enter dangerous User Agent territory.

denschub:
Let's go with #22 instead.

PR #22, opened 2018-04-13
Bug 1452707 - Shim window.controllers for ib.absa.co.za

denschub:
This deprecates #21 as per [this comment](https://bugzilla.mozilla.org/show_bug.cgi?id=1452707#c5).

r? @wisniewskit for the technical part,
r? @miketaylr as a second reviewer and management approval!

### DIFF
--- a/src/webextension/injections/js/bug1452707-window.controllers-shim-ib.absa.co.za.js
+++ b/src/webextension/injections/js/bug1452707-window.controllers-shim-ib.absa.co.za.js
@@ -1,0 +1,25 @@
+"use strict";
+
+/**
+ * Bug 1452707 - Build site patch for ib.absa.co.za
+ * WebCompat issue #16401 - https://webcompat.com/issues/16401
+ *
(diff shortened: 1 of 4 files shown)


src/webextension/injections/js/bug1452707-window.controllers-shim-ib.absa.co.za.js
```diff
@@ -0,0 +1,25 @@
...
+ *
+ * This site patch simply sets window.controllers to a string, resulting in
+ * their check to work again.
+ */
+
+/* globals exportFunction */
+
+Object.defineProperty(window.wrappedJSObject, "controllers", {
+  get: exportFunction(function() {
+    return "window.controllers has been shimmed with a string to get this site's browser detection to work in Firefox 61+.";
```
miketaylr: +1 to this approach, seems much safer. I don't know if controllers will be removed in 61 though, so this comment might need to be made more generic or updated once we know what the actual train is.

PR #23, opened 2018-06-29
Bug 1457335 - Override UA and navigator.vendor for histography.io

denschub:
See https://bugzilla.mozilla.org/show_bug.cgi?id=1457335 and linked bugs for details.

r? @wisniewskit

### DIFF
--- a/src/webextension/injections/js/bug1457335-histography.io-ua-change.js
+++ b/src/webextension/injections/js/bug1457335-histography.io-ua-change.js
@@ -1,0 +1,30 @@
+"use strict";
+
+/**
+ * Bug 1457335 - histography.io - Override UA & navigator.vendor
+ * WebCompat issue #1804 - https://webcompat.com/issues/1804
+ *
+ * This site is using a strict matching of navigator.userAgent and
+ * navigator.vendor to allow access for Safari or Chrome. Here, we set the
+ * values appropriately so we get recognized as Chrome.
+ */
+
+/* globals exportFunction */
+
+const CHROME_UA = navigator.userAgent + " Chrome for WebCompat";
+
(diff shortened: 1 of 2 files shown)


src/webextension/injections/js/bug1457335-histography.io-ua-change.js
```diff
@@ -0,0 +1,30 @@
...
+ * navigator.vendor to allow access for Safari or Chrome. Here, we set the
+ * values appropriately so we get recognized as Chrome.
+ */
+
+/* globals exportFunction */
+
+const DEFAULT_UA = navigator.userAgent;
+
+Object.defineProperty(window.wrappedJSObject.navigator, "userAgent", {
+  get: exportFunction(function() {
```
wisniewskit: It might be worth caching this into a `const` like in the BoA patch.

src/webextension/injections/js/bug1457335-histography.io-ua-change.js
```diff
@@ -0,0 +1,30 @@
...
+    return DEFAULT_UA + " Chrome for WebCompat";
+  }, window),
+
+  set: exportFunction(function() {}, window)
+});
+
+Object.defineProperty(window.wrappedJSObject.navigator, "vendor", {
+  get: exportFunction(function() {
```
wisniewskit: Tiniest of nits here, but I Chrome uses "Google Inc." with the period at the end. I suspect it doesn't matter at all, but I thought I'd mention it.

denschub:
Although it's just a simple string concat here, you are right, we can cache that as well. Done.

Also thanks for the `navigator.vendor` catch, although it's irrelevant here, I corrected it just in case. :)

PR #24, opened 2018-06-29
Bug 1472075 - UA override for Bank of America for OSX & Linux

denschub:
See https://bugzilla.mozilla.org/show_bug.cgi?id=1472075 and linked bugs for details.

r? @wisniewskit

### DIFF
--- a/src/webextension/injections/js/bug1472075-bankofamerica.com-ua-change.js
+++ b/src/webextension/injections/js/bug1472075-bankofamerica.com-ua-change.js
@@ -1,0 +1,41 @@
+"use strict";
+
+/**
+ * Bug 1472075 - Build UA override for Bank of America for OSX & Linux
+ * WebCompat issue #2787 - https://webcompat.com/issues/2787
+ *
+ * BoA is showing a red warning to Linux and macOS users, while accepting
+ * Windows users without warning. From our side, there is no difference here
+ * and we receive a lot of user complains about the warnings, so we spoof
+ * as Firefox on Windows in those cases.
+ */
+
+/* globals exportFunction */
+
+if (!navigator.platform.includes("Win") && !navigator.userAgent.includes("Android")) {
+  const WINDOWS_UA = navigator.userAgent.replace(/\(.*; rv:/i, "(Windows NT 10.0; Win64; x64; rv:");
+
+  Object.defineProperty(window.wrappedJSObject.navigator, "userAgent", {
+    get: exportFunction(function() {
+      return WINDOWS_UA;
+    }, window),
+
+    set: exportFunction(function() {}, window)
+  });
+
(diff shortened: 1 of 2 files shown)


src/webextension/injections/js/bug1472075-bankofamerica.com-ua-change.js
```diff
@@ -0,0 +1,41 @@
...
+ *
+ * BoA is showing a red warning to Linux and macOS users, while accepting
+ * Windows users without warning. From our side, there is no difference here
+ * and we receive a lot of user complains about the warnings, so we spoof
+ * as Firefox on Windows in those cases.
+ */
+
+/* globals exportFunction */
+
+if (!navigator.platform.includes("Win")) {
```
wisniewskit: This will cover Fennec as well, no? Do we want it to, when Fennec supports GoFaster? If not we might want to also add something like `&& navigator.userAgent.includes("Android")` just to future-proof it a bit.

src/webextension/injections/js/bug1472075-bankofamerica.com-ua-change.js
```diff
@@ -0,0 +1,41 @@
...
+  const DEFAULT_UA = navigator.userAgent;
+
+  Object.defineProperty(window.wrappedJSObject.navigator, "userAgent", {
+    get: exportFunction(function() {
+      return DEFAULT_UA.replace(/\(.*; rv:/i, "(Windows NT 10.0; Win64; x64; rv:");
```
wisniewskit: If the site reads `userAgent` a lot, it might be worth pulling this out into its own `const` like `REPLACED_UA` for a (very minor) perf optimization.

denschub:
Thanks for the feedback! Indeed, making sure we don't override on Android is a good point, and I also cached the UA rewrite in a const. :)

dholbert:
@denschub Looks like this has been sitting for a month -- were we still waiting on anything before landing this?  Just want to make sure it didn't drift off of the radar right when it was (maybe) ready. :)

denschub:
@dholbert Thanks for the ping! :)

I am currently working on getting this extension running on Fennec, as we have site patches waiting for Fennec as well. I am almost done there (in fact, preparing the patches right now).

We want to keep the codebase identical between Fenenc and Desktop, so I waited with landing the new patches on Desktop until we are sure the new code (there have been some changes besides just adding site patches) works on Fennec as well, so we can ship the new version with actual site patches to both Fennec **and** Desktop at the same time.

In hindsight, and as something to learn for the future, I shouldn't have done this. We had some issues with Fennec, which delayed landing the patches for quite a long time, but at some point, we should have decided to push the Desktop update anyway, regardless of what's up with Fennec.

We are still kinda figuring out the actual process behind this, but this is something we shouldn't repeat, for sure.

PR #25, opened 2018-06-29
Bug 1472081 - election.gov.np - Override window.sidebar with something falsey

denschub:
See https://bugzilla.mozilla.org/show_bug.cgi?id=1472081 and linked bugs for details.

r? @wisniewskit

### DIFF
--- a/src/webextension/background.js
+++ b/src/webextension/background.js
@@ -10,6 +10,12 @@ const contentScripts = [
     css: [{file: "injections/css/bug0000000-dummy-css-injection.css"}],
     js: [{file: "injections/js/bug0000000-dummy-js-injection.js"}],
     runAt: "document_start"
+  },
+  {
+    matches: ["http://202.166.205.141/bbvrs/*"],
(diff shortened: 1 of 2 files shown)


src/webextension/background.js
```diff
@@ -10,6 +10,12 @@ const contentScripts = [
     css: [{file: "injections/css/bug0000000-dummy-css-injection.css"}],
     js: [{file: "injections/js/bug0000000-dummy-js-injection.js"}],
     runAt: "document_start"
+  },
+  {
+    matches: ["http://202.166.205.141/*"],
```
wisniewskit: Are we sure this is a static IP?

Also, whois is telling me it's worldlink.com, not gov.np, so I'm not sure if there's a chance this would affect other sites as well.

denschub: I am not 100% certain, but fairly, as it seems to be unchanged for at least 8 moths, see https://github.com/webcompat/web-bugs/issues/11622#issuecomment-334643852.

You are right about maybe being too eager here, and it should be safe to target only `http://202.166.205.141/bbvrs/`. I'll run a test with it and update if that is good enough.

PR #26, opened 2018-08-03
No bug - Add support for platform-specific content scripts

denschub:
This PR enables us to write platform-specific content scripts, as well as universal ones. This is needed to avoid Fennec running overrides that are not designed for Fennec.

r? @wisniewskit

### DIFF
--- a/src/webextension/background.js
+++ b/src/webextension/background.js
@@ -4,43 +4,55 @@
  *
  * https://wiki.mozilla.org/Compatibility/Go_Faster_Addon/Override_Policies_and_Workflows#User_Agent_overrides
  */
-const contentScripts = [
-  {
-    matches: ["*://webcompat-addon-testcases.schub.io/*"],
-    css: [{file: "injections/css/bug0000000-dummy-css-injection.css"}],
-    js: [{file: "injections/js/bug0000000-dummy-js-injection.js"}],
-    runAt: "document_start"
-  },
-  {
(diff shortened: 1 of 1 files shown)


src/webextension/background.js
```diff
@@ -4,43 +4,55 @@
...
 
 let port = browser.runtime.connect();
 let registeredContentScripts = [];
 
-function registerContentScripts() {
-  contentScripts.forEach(async (contentScript) => {
+async function registerContentScripts() {
+  let platform = "desktop";
+  let platformInfo = await browser.runtime.getPlatformInfo();
+  if (platformInfo.os == "android") {
```
wisniewskit: This is fine, but I'm wondering if we'll need to consider GeckoView separately from Fennec at some later date. Perhaps using "android" instead of "fennec" would make it clearer for future readers that overrides should cover both browsers?

denschub: Hm, that's a good point for the future, I was thinking about that as well. Right now, I decided to stick with `fennec`, as that is what we are using in `ua_overrides.jsm`. There, I have access to `Services.appinfo.name.toLowerCase()`, which returns `fennec`, and I wanted to keep it consistent.

You are right, though, this might also catch GeckoView products, or any other products on Android, if they ever receive extensions.

denschub: Renamed it to "android" instead of "fenenc". No-surprises is better than having the same identifier in two places, especially since they technically don't do the same thing. Thanks. :)

PR #33, opened 2018-08-30
PoC for shimming ::-webkit-scrollbar

upsuper:
This PR is **not** intended to be merged.

It's just a proof of concept to show a possible way to shim `::-webkit-scrollbar` support with the new scrollbar properties. It should even be possible to simulate the whole webkit scrollbar stuff to some extend (requiring inserting more elements and script into the document). This can be done if necessary, but I tend not to do this at all.

(Note that this currently requires enabling both `layout.css.scrollbar-width.enabled` and `layout.css.unknown-webkit-pseudo-element` on Nightly to take effect.)

Once we ship the scrollbar properties, it shouldn't be too hard for websites to switch for some common usecases (coloring scrollbar or having thinner scrollbar). For more advanced usecases, it should be easy to build encapsulated component to do. So hopefully we would never need something like this.

### DIFF
--- a/src/webextension/background.js
+++ b/src/webextension/background.js
@@ -40,7 +40,12 @@ const contentScripts = {
       js: [{file: "injections/js/bug1482066-portalminasnet.com-window.sidebar-shim.js"}],
       runAt: "document_start",
       allFrames: true
-    }
+    },
+    {
+      matches: ["https://gitter.im/*"],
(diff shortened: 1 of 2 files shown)


upsuper:
FWIW, this PoC would fix webcompat/web-bugs#12356, but it doesn't seem to me other issues linked to bug 1432935 can be fixed with this.

denschub:
Let's continue this on a rebased version in #45.

PR #35, opened 2018-09-27
Port to WebExtension

denschub:
r? @wisniewskit


denschub:
Patch has an r+ from aswan in [bug 1498278](https://bugzilla.mozilla.org/show_bug.cgi?id=1498278), and the patch is awaiting checkin. so let's just merge this.

@wisniewskit if you have any remarks, we can always do a follow-up pr, so please feel free. :)

PR #39, opened 2018-11-26
Bug 1505828 - Add further UA overrides

denschub:
See individual commits for bug numbers.


denschub:
I thought about this, especially since we're now using the same code multiple times. I don't feel comfortable just doing that every time and passing it as a second argument into the UA transformer, but maybe adding a little object containing some helper functions somewhere doesn't hurt?

wisniewskit:
Would it not work to just add another variable above UAOverrides in `ua_overrides.js`, say OriginalUAPrefix, and access it from inside the arrow functions via regular closure? That might be good enough.

denschub:
Updated the PR to include a reusable function `UAHelpers.getPrefix` as discussed on IRC.

denschub:
r+ via IRC, thanks for the fast reviews. :)

PR #45, opened 2019-02-12
Bug XXX - Add shim for -webkit-scrollbar

denschub:
**Not ready to merge**, only as a base for further discussions.


denschub:
Excuse me, that was an accident.

miketaylr:
@denschub we can probably close this now, right?

denschub:
Yeah. We decided against this and shipped simple CSS injections instead.

PR #48, opened 2019-03-15
Multiple bugs - Address two ::-webkit-scrollbar cases

denschub:
As a Proof-of-Concept, let's address two cases of broken layouts caused by the use of `::-webkit-scrollbar`. This PR contains interventions for two sites:

* [Bug 1305028](https://bugzilla.mozilla.org/show_bug.cgi?id=1305028): Here, the intervention hides the scrollbar in the main player section only. YouTube is using `::-webkit-scrollbar` to make all scrollbars on the site a fixed width, but as far as I can tell, only the main player area scrollbar is an issue, because it gets partly overlapped by the video player.
* [Bug 1518781](https://bugzilla.mozilla.org/show_bug.cgi?id=1518781): The scrollbar prevents users to click on the three-dot-button to access chat message interactions reliably. As discussed in scrollbar, setting the width to `thin` is not enough, so this intervention removes the scroll bar there.

r? @wisniewskit

### DIFF
--- a/src/injections.js
+++ b/src/injections.js
@@ -93,6 +93,15 @@ for (const injection of [
       matches: ["*://*.twitch.tv/*"],
       css: [{file: "injections/css/bug1518781-twitch.tv-webkit-scrollbar.css"}],
     },
+  }, {
+    id: "bug1305028",
+    platform: "desktop",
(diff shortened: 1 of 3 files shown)


miketaylr:
+1 for simplicity

PR #52, opened 2019-05-29
Roll-up of about:compat QA fixes and the PDK5 intervention. Fixes #43

### LINKED ISSUE #43: PDK5 sites: Stay with the whitelist approach or build something that detects CDN loads?
@wisniewskit can you send in your POC as a PR, or link it here?

It would be good to try to find some other sites using PDK5 (perhaps with HTTPArchive or similar?), and do testing on say 5 of them. If it seems OK, let's go the CDN route.

wisniewskit:
Sorry for the delay here. This is a roll-up patch containing the last batch of fixes for about:compat, as well as the PDK5 intervention (this makes us match what's in mozilla-central right now).


denschub:
Thanks. <3

PR #53, opened 2019-06-11
Bug 1432935 - Add remaining -webkit-sidebar cases

denschub:
Only two left! All the other sites fixed themselves after we were strong with outreach. :)

r? @wisniewskit

### DIFF
--- a/src/injections/css/bug1432935-discordapp.com-webkit-scorllbar-white-line.css
+++ b/src/injections/css/bug1432935-discordapp.com-webkit-scorllbar-white-line.css
@@ -1,0 +1,16 @@
+/**
+ * discordapp.com - -webkit-scrollbar dependency causes visible white line
+ * Part of Bug #1432935 - https://bugzilla.mozilla.org/show_bug.cgi?id=1432935
+ * WebCompat issue #7919 - https://webcompat.com/issues/7919
+ *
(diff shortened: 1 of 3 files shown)


src/injections/css/bug1432935-discordapp.com-webkit-scorllbar-white-line.css
```diff
@@ -0,0 +1,16 @@
...
+ * WebCompat issue #7919 - https://webcompat.com/issues/7919
+ *
+ * Discord depends on -webkit-scrollbar for styling and hiding their scrollbars
+ * in the UI. Previously, the scrollbars overlapped content permanently.
+ * However, this issue seems to be addressed now, but a small white line
+ * still remains. While Discord is working on this, let's get rid of these
+ * lines.
+ */
+.themeGhostHairline-DBD-2d .pad-29zQak,
+.themeGhostHairlineChannels-3G0x9_ .pad-29zQak {
```
wisniewskit: We may need to monitor this one, since it looks like the class names will change if they change that CSS file (even if they don't fix the scrollbar issue).

miketaylr:
Can we get this landed and into m-c @denschub?

denschub:
@miketaylr Yeah. Will check with Ksenia how the removal of no longer used interventions is going, if those are close, having both bundled together probably makes a lot of sense. :)

PR #56, opened 2019-06-28
Bug 1561371 - allow GMail to scroll horizontally on Fennec

wisniewskit:


src/injections/css/bug1561371-mail.google.com-allow-horizontal-scrolling.css
```diff
@@ -0,0 +1,3 @@
+body > #views > div {
```
denschub: This is looking good, but in the other interventions, I [added a little comment block](https://github.com/mozilla/webcompat-addon/blob/499e369259db96957232751acc0479d2506fe07a/src/injections/css/bug1518781-twitch.tv-webkit-scrollbar.css) explaining what and why is done here. I don't think there is harm to that, especially since those source files can be visible in the devtools without their context in the addon. Do you mind adding a comment block like here as well?

PR #57, opened 2019-07-08
Bug 1444192 - Replace static testing interventions with automated tests

denschub:
This PR probably looks a bit larger than it actually is, and that's because I had to refactor large portions of the code to make our plans work. I absolutely recommend to look at the individual commits instead of the whole PR at once, because there is a lot of information in the commit messages, and it should make things a lot easier to understand. I hope.

The main goal here was to replace the static test cases hosted on my personal servers with some automated tests, so that's what this PR does. However, our little addon has grown up a bit recently, and it got to a point where the code wasn't really testable as is. The recent addition of `about:compat` made things harder for me, and it's ultimately the reason why this PR is significantly delayed because I had to refactor my own refactoring several times. :) The thing is that `about:compat` needs to know which interventions are currently active, and it also needs to communicate back to disable things. We've used global variables previously, but those make testing basically impossible, so I had to find another solution here.

We deliberately decided against using mochitest or other in-tree frameworks to test this extension. The main reason for this is the fact that this extension is not an in-tree-only extension. Right now, we are actively working on getting it included in android-components so that Fenix and future products can benefit from our work as well, so having in-tree tests don't quite cut it. Also, spinning up a Firefox build to run tests for this repo would take significantly longer. So, for now, this project is tested by some standalone, Node.js-based unit tests. If we have additional testing requirements in the future, we can add those, but I do feel this is a good start.

In the process of working on this, I made some bold decisions to get things done. They will become apparent when looking at the commit messages, and I also explained my reasoning behind them, but I want to highlight those in this PR text as well, so you can spot them easily:

* I pretty much refactored all the things. We previously had a flat file hierarchy and just had the JS files do things when they needed to do. I changed that. Besides introducing what I believe is a more neat directory structure, I also encapsulated all the logic into their own classes and added a `run.js` that ultimately ends up instantiating those classes and running things in the browser. This was done for both maintainability, but also for making things testable. Testing JS files that just call functions and do things is almost impossible to do cleanly. I do believe that these changes improve the project, but I am biased because I made those changes. If you feel like this is a step backward, please make some noise, so we can collaborate on finding a solution that makes everyone happy.
* These refactorings move `aboutCompat.html` into a new directory. @wisniewskit, what's the impact for l10n here?
* I removed eslint support for this project. The reasons are explained in the commit, but the tl;dr is: If we export our sources into m-c anyway, we can just use `mach lint` for our code style checks, which removes the need to maintain the rules ourselves.
* To make things testable, I had to modularize the code. Since we cannot use JS modules inside webextensions yet, the code now includes a somewhat double-barreled approach. For running in the browser, all files just get included one after another and run in a global scope. For testing, I added `module.exports` statements so I can `require()` those modules in our tests. However, these `module.exports` statements will be included when exporting into m-c, and there, they could actually cause an exception, because `module` will not be defined. That's the reason behind `src/lib/module_shim.js`: it makes those export statements not fail in the browser. Ultimately, I believe those are cheap as they just keep a reference to the class that exists anyway, but there is a chance we might not be able to land this. If that's the case, we can remove the `module.exports` lines via jake, but I decided to try the simpler approach first.

Ultimately, the test suite I ended up with was more straightforward than I originally wrote, but there are some good reasons for that, and it boils down to great design decisions when we decided to write this extension (yes, I'm patting my own back here a bit):

* Most of the heavy lifting happens inside the webextension APIs. We don't need to test those, we have lots of tests in the tree.
* Our "config files" for injections are straight-up WebExtenstion contentScript configs. There is no complex abstraction layer, we write things and they get passed directly into the WebExtension API. No magic.
* However, this allows me to run tests on the actual definitions themselves, instead of writing complex mockings for the APIs. For example, I test if the contentScript files are actually there, if we accidentally used a duplicate ID (we had issues with `about:compat` breaking because of that before), and generally do some sanity-checks on those files. I do quite like that.

There are some things that, given the decisions we made, we cannot test: everything that needs to be tested inside the browser, basically. So this is `about:compat`, and the actual interventions. For `about:compat`, involving QA seems to be absolutely fine, given the scope is minimal, and they always discover fascinating bugs. For the interventions themselves, this was decided to be totally out of scope. For now, as the number is still low, we are fine with having our Softvision friends manually check these once in a while. The test scripts for those would be really complicated, given you need logins or complicated steps to reproduce for most of them, so having those tested automatically is nothing we can quickly achieve. I also was unable to write tests for the streaming replace function used for PDK5 sites, because I'd have to fake a complete stream. I don't think that's a dealbreaker, and I'm sure we can improve on that in the future.

Also, this test suite is obviously not final. This PR does a lot of groundwork, so adding more tests in the future should be a lot simpler. It's an improvement, but absolutely not the end goal.

r? @wisniewskit
r? @miketaylr (that's probably just a feedback?, but I want to trick your mail filters!)

If you, or anyone else, has any questions, ideas, concerns, improvements, wishes, ... please do let me know.

Also, sorry for this wall of text, but I felt like I had to explain some things, as I could not assume that everyone thinks the same way I do. :) For the record, I did run in-tree tests, which did not fail, and also did run Talos, which did not show significant changes, so I guess that I didn't do horrible things.

---

```
# npm test

> webcompat-gofaster@1.0.0 test /Users/denschub/Work/gofaster
> jasmine

Randomized with seed 42443
Started
.......................................................................................................................................................


151 specs, 0 failures
Finished in 0.101 seconds
Randomized with seed 42443 (jasmine --random=true --seed=42443)
```

:)


wisniewskit:
>what's the impact for l10n here?

I think we'll actually be fine, as we're using ftl in a way that doesn't seem will be broken by such a change. We might want to ping @flod just in case, though.

src/lib/ua_overrides.js
```diff
@@ -0,0 +1,119 @@
...
+  }
+
+  enableOverride(override) {
+    if (override.active) {
+      return;
+    }
+
+    const { matches, uaTransformer } = override.config;
+    const listener = details => {
+      for (var header of details.requestHeaders) {
```
wisniewskit: might as well change `var` to `let` here

spec/helpers/mock_about_compat_brorker.js
```diff
@@ -0,0 +1,12 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
```
wisniewskit: brorker again (filename)

spec/available_injections.spec.js
```diff
@@ -0,0 +1,71 @@
...
+describe("AVAILABLE_INJECTIONS", () => {
+  it("contais at least one injection definition", () => {
+    expect(AVAILABLE_INJECTIONS.length).toBeGreaterThanOrEqual(1);
+  });
+
+  it("provide a unique ID for each injection", () => {
+    let ids = new Set();
+    let duplicates = new Set();
+
+    for (let injection of AVAILABLE_INJECTIONS) {
```
wisniewskit: extreeeeeeme nit!  feel free to use `const` instead of `let` in these kinds of loops (if you're hardcore enough to go that far :)

src/about-compat/aboutCompat.js
```diff
@@ -26,15 +26,21 @@ const portToAddon = (function() {
     return Promise.reject("background script port disconnected");
   }
 
-  return {send};
-}());
+  return { send };
+})();
```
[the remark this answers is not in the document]
denschub: `¯\_(ツ)_/¯ `

denschub:
Nothing annoying at all! All fixed.

miketaylr:
f+, thanks for the detailed blog post i mean commit summary.

PR #62, opened 2019-07-17
Add testbed interventions for manual verification.

denschub:
This re-adds the manual verification testbed as we've discussed yesterday. It points to [the new testbed](https://github.com/mozilla/webcompat-addon-testbed/), which is auto-deployed and can be managed by the entire team, so there no longer is a dependency or a reference to my private infrastructure in there. 🎉 

My commit is complete, but I'd like this PR to contain a second commit that blacklists the test interventions from `about:config`. @wisniewskit, I pushed this as a branch on this repo, `testbed-interventions`, please feel free to push your commit doing the blacklisting into the branch whenever you have time to do so. :)

### DIFF
--- a/src/about-compat/aboutCompat.js
+++ b/src/about-compat/aboutCompat.js
@@ -6,6 +6,9 @@
 
 /* globals browser */
 
+let availablePatches;
+let showHidden = false;
+
 const portToAddon = (function() {
   let port;
 
@@ -69,7 +72,8 @@ Promise.all([
     }
   });
 
-  redraw(info);
+  availablePatches = info;
+  redraw();
 });
 
 function onMessageFromAddon(msg) {
@@ -94,13 +98,17 @@ function onMessageFromAddon(msg) {
   button.disabled = !!msg.toggling;
 }
 
-function redraw(info) {
-  const { overrides, interventions } = info;
-  redrawTable($("#overrides"), overrides);
-  redrawTable($("#interventions"), interventions);
+function redraw() {
+  if (!availablePatches) {
+    return;
+  }
+  const { overrides, interventions } = availablePatches;
+  const showHidden = location.hash === "#all";
+  redrawTable($("#overrides"), overrides, showHidden);
+  redrawTable($("#interventions"), interventions, showHidden);
 }
 
-function redrawTable(table, data) {
+function redrawTable(table, data, showHidden = false) {
   const df = document.createDocumentFragment();
   table.querySelectorAll("tr").forEach(tr => {
     tr.remove();
@@ -128,6 +136,10 @@ function redrawTable(table, data) {
   }
 
   for (const row of data) {
+    if (row.hidden && !showHidden) {
+      continue;
+    }
+
     const tr = document.createElement("tr");
     tr.setAttribute("data-id", row.id);
     df.appendChild(tr);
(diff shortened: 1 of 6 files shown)


src/data/ua_overrides.js
```diff
@@ -13,22 +13,37 @@
...
   {
     /*
-    * Bug 1563839 - rolb.santanderbank.com - Build UA override
-    * WebCompat issue #33462 - https://webcompat.com/issues/33462
-    *
-    * santanderbank expects UA to have 'like Gecko', otherwise it runs
-    * xmlDoc.onload whose support has been dropped. It results in missing labels in forms
-    * and some other issues.  Adding 'like Gecko' fixes those issues.
-    */
+     * Bug 1563839 - rolb.santanderbank.com - Build UA override
```
denschub: Prettier made me do that.

denschub:
I was about to r+ and merge this, because it works just fine. However, I've noticed something.

With us hiding the testbed interventions from `about:compat`, there is also no way to test if disabling things on `about:compat` actually works... @wisniewskit, how difficult would it be to skip hiding if, let's say, we access `about:compat#all` or something like that?

miketaylr:
> there is also no way to test if disabling things on `about:compat` actually works...

Do we have automated browser tests that cover this?

denschub:
@miketaylr we do have tests that check if the right methods get called when we send the right messages - but there is no way to actually testing about:compat itself, without building a browser alongside the extension and doing behavioral tests within that browser. So the answer is "no", because we can't really test `about:compat` with non-in-tree tests.

wisniewskit:
Another option would be for us to just output mochitests, and then let the m-c CI do the job?

src/about-compat/aboutCompat.js
```diff
@@ -6,6 +6,13 @@
 
 /* globals browser */
 
+let availablePatches;
+let showHidden = false;
+function checkHash() {
+  redraw();
+}
+window.onhashchange = checkHash;
```
denschub: This appears to be the only use of `checkHash`, is there something stopping us from doing

```js
window.onhashchange = redraw;
```
?

miketaylr:
> Another option would be for us to just output mochitests, and then let the m-c CI do the job?

Yeah, I guess that was my question, by "browser tests" I meant mochitest browser tests. It seems like it's possible to open up a tab to about:compat, click on a button, open a new tab and read a pixel value or HTML title, or whatever. But it sounds like we don't have those now -- is the real desire to have the ability to do a quick sanity check during dev? Otherwise the tests would catch any regressions in Try.

denschub:
> is the real desire to have the ability to do a quick sanity check during dev

That was the intention I got. And for QA in case we need manual verification.

I'm obviously not opposed of having mochitests in tree, we'd just find a way to maintain those (like, would those just be committed directly to m-c? Or part of this repo? How would that all look like?)

miketaylr:
> like, would those just be committed directly to m-c

That probably makes sense. Just "sanity tests" to make sure other people to break out stuff (it's always other people, never us!).

But we can kick the can down the road and not block progress on this.

src/about-compat/aboutCompat.js
wisniewskit: Good catch, I'll update now.

denschub:
r+ from me for Tom's changes, and Tom r+'ed my changes earlier as well. CI is failing, but that's because the Docker Registry is down. 🤷‍♂

PR #64, opened 2019-07-18
Emit a console.warn when doing any overrides on a tab. Fixes #47

### LINKED ISSUE #47: Emit a console.warn when doing an override
In https://bugzilla.mozilla.org/show_bug.cgi?id=1531509 it was reported that having a console message could help developer understand what's going on.

A console warning could be a good start (the message can be styled to be bigger than other messages if we want that `console.warn("%cThis website has compatibility issues […]", "font-size: 48px;")`).

wisniewskit:
Note that no attempt is made in this patch to do any l10n on the related messages, because there is no FTL interface on content scripts for system addons yet (and it can rely on strings which already aren't being localized, like the ones for the PDK fix, which is hard-coded to use the string "Sites using PDK 5" right now about:compat).

Unless I've missed a memo or two, we currently would have to do things how the [webcompat reporter is doing](https://searchfox.org/mozilla-central/source/browser/extensions/report-site-issue/experimentalAPIs/l10n.js) (which I think other addons are doing as well). But I'm not sure if that's what we want to do. We should check with the l10n team to make a decision here.

### DIFF
--- a/src/lib/console_warning_helper.js
+++ b/src/lib/console_warning_helper.js
@@ -1,3 +1,11 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+"use strict";
+
+/* globals module */
+
 const ConsoleWarningEyeCatch = "Oh no!";
 const ConsoleWarningDetails = `
 This web site has a web compatibility issue.
@@ -18,3 +26,5 @@ function promiseConsoleWarningScript(domain) {
            }`,
   });
 }
+
+module.exports = promiseConsoleWarningScript;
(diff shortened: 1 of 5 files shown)


wisniewskit:
Yes, I'll squash the commits/etc after I've had a chance to consider l10n, just in case we should adopt the same solution other system addons are using for now (which wouldn't be too tricky to add to this patch).

miketaylr:
What's the status here @wisniewskit?

miketaylr:
@denschub is this something we still want, or should we close this?

denschub:
@miketaylr I'd say this is a solid "nice to have", still. Probably ignoring l10n and just having it English-only is better than not emitting a warning in some cases.

miketaylr:
@denschub thanks, can you take this over then?

denschub:
Self-assigning this, I'll do the rebase. :)

denschub:
Rebased in #185.

PR #69, opened 2019-07-23
Bug 1567945 - override for beeg.com and Bug 1568256 - intervention for zertifikate.commerzbank.de 

ksy36:
r? @denschub

### DIFF
--- a/src/data/injections.js
+++ b/src/data/injections.js
@@ -225,6 +225,20 @@ const AVAILABLE_INJECTIONS = [
       ],
     },
   },
+  {
+    id: "bug1568256",
+    platform: "android",
+    domain: "zertifikate.commerzbank.de",
+    bug: "1568256",
+    contentScripts: {
+      matches: ["*://*.zertifikate.commerzbank.de/webforms/mobile/*"],
+      css: [
+        {
+          file: "injections/css/bug1568256-zertifikate.commerzbank.de-flex.css",
+        },
+      ],
+    },
+  },
 ];
 
 module.exports = AVAILABLE_INJECTIONS;
--- a/src/data/injections.js
+++ b/src/data/injections.js
@@ -213,7 +213,7 @@ const AVAILABLE_INJECTIONS = [
   },
   {
     id: "bug1567610",
-    platform: "desktop",
+    platform: "all",
     domain: "dns.google.com",
     bug: "1567610",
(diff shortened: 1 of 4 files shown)


src/data/ua_overrides.js
```diff
@@ -287,7 +287,7 @@ const AVAILABLE_UA_OVERRIDES = [
     domain: "posts.google.com",
     bug: "1566253",
     config: {
-      matches: ["*://posts.google.com/share/*"],
+      matches: ["*://posts.google.com/*"],
       uaTransformer: _ => {
```
ksy36: This override worked on the original reported page, but then when clicking on "More" and trying to load more posts here https://posts.google.com/share/iT2YoKXw?sa=X&ved=0ahUKEwi7nqT01cvjAhVsg-AKHS12BQoQrpcBCAw ,
the response won't return anything :( So I've added the full domain instead.

src/data/injections.js
```diff
@@ -213,7 +213,7 @@ const AVAILABLE_INJECTIONS = [
   },
   {
     id: "bug1567610",
-    platform: "desktop",
+    platform: "all",
```
wisniewskit: Was this also an intended change for this patch?

src/injections/css/bug1568256-zertifikate.commerzbank.de-flex.css
```diff
@@ -0,0 +1,12 @@
...
+ * zertifikate.commerzbank.de - clickable elements on the page are collapsed
+ * Bug #1568256 - https://bugzilla.mozilla.org/show_bug.cgi?id=1568256
+ * WebCompat issue #9102 - https://webcompat.com/issues/9102
+ *
+ * Affected elements have display:-webkit-box and display:flex applied, however,
+ * listed in wrong order, so display:-webkit-box is becoming the final say.
+ * Adding display: flex for those elements fixes the issue
+ */
+.x-layout-box {
+    display: flex!important;
```
denschub: Nit: Two spaces instead of four, space before `!important`. (That should have also applied to the google DNS override, but I missed that one, so let's let that slide.)

src/data/injections.js
ksy36: yeah, I was able to test this properly after building Firefox on Android, so the fix works fine

PR #72, opened 2019-07-25
Change the codebase-principal fix for bug 1560455 to support both old…

### LINKED ISSUE #71: The codebase principal fix needs to be tweaked to support Firefox 68 and 69
The fix we made in commit 3b128b130a2bfcc8ae6d9440d64eb6f71425453d needs to handle both function names so it doesn't break on the current Firefox release (the change didn't land in 68, it will make it to 69).

wisniewskit:
… and new function names. Fixes #71

Note that I've already updated our proposed unsigned XPI on bug 1568636.


denschub:
Merged as https://github.com/mozilla/webcompat-addon/commit/aa92cc03cfb223ec1b9d1dbf063e4e599d63b6be with changed commit message as discussed on IRC.

PR #75, opened 2019-08-06
Bug 1570390 - UA override for zhidao.baidu.com, Bug 1570119 - Intervention for teamcoco.com and Bug 1570328 - Intervention for developer.apple.com

ksy36:
r? @denschub

### DIFF
--- a/src/injections/css/bug1570119-teamcoco.com-scrollbar-width.css
+++ b/src/injections/css/bug1570119-teamcoco.com-scrollbar-width.css
@@ -1,0 +1,11 @@
+/**
+ * teamcoco.com - a scrollbar at the top covering navigation menu
+ * Bug #1570119 - https://bugzilla.mozilla.org/show_bug.cgi?id=1570119
+ *
+ * The scrollbar is covering navigation items making them unusable.
+ * There are ::-webkit-scrollbar css rules already applied to the scrollbar,
+ * hiding it in Chrome. Adding the scrollbar-width: none fixes the issue in Firefox.
+ */
+.css-bdnz85 {
+  scrollbar-width: none;
+}
--- a/src/injections/css/bug1570328-developer-apple.com-transform-scale.css
+++ b/src/injections/css/bug1570328-developer-apple.com-transform-scale.css
@@ -1,0 +1,17 @@
+/**
+ * developer.apple.com - content of the page is shifted to the left
+ * Bug #1570328 - https://bugzilla.mozilla.org/show_bug.cgi?id=1570328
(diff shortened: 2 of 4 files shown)


src/injections/css/bug1570119-teamcoco.com-scrollbar-width.css
```diff
@@ -0,0 +1,11 @@
...
+ * teamcoco.com - a scrollbar at the top covering navigation menu
+ * Bug #1570119 - https://bugzilla.mozilla.org/show_bug.cgi?id=1570119
+ *
+ * The scrollbar is covering navigation items making them unusable.
+ * There are ::-webkit-scrollbar css rules already applied to the scrollbar,
+ * hiding it in Chrome. Adding the scrollbar-width: none fixes the issue in Firefox.
+ */
+.css-bdnz85 {
+    scrollbar-width: none;
+}
```
ksy36: Forgot to run prettier, it's fixed in the next commit.

src/injections/css/bug1570328-developer-apple.com-transform-scale.css
```diff
@@ -0,0 +1,18 @@
...
+ * Bug #1570328 - https://bugzilla.mozilla.org/show_bug.cgi?id=1570328
+ * WebCompat issue #4070 - https://webcompat.com/issues/4070
+ *
+ * The site is relying on zoom property which is not supported by Mozilla,
+ * see https://bugzilla.mozilla.org/show_bug.cgi?id=390936. Adding a combination
+ * of transform: scale(1.4), transform-origin and width fixes the issue
+ */
+@media only screen and (min-device-width: 320px) and (max-device-width: 980px),
+  (min-device-width: 1024px) and (max-device-width: 1024px) and (min-device-height: 1366px) and (max-device-height: 1366px) and (min-width: 320px) and (max-width: 980px),
+  (min-device-width: 1024px) and (max-device-width: 1024px) and (min-device-height: 1366px) and(max-device-height: 1366px) and (min-width: 320px) and (max-width: 980px) {
```
denschub: I have no idea why prettier does not care about this, but `and(max-device-height: 1366px)` is missing a space after the `and`.

denschub: so we even need line 12? Isn't it exactly the same as the previous line? What am I missing? :)

ksy36: Ah, good point. I've just copied it from the website's style sheet. I'll remove the last line

PR #82, opened 2019-09-20
Bug 1582582 - UA override for sling.com

ksy36:
r? @denschub

### DIFF
--- a/src/manifest.json
+++ b/src/manifest.json
@@ -30,7 +30,7 @@
     }
   },
 
-  "content_security_policy": "script-src 'self' 'sha256-MmZkN2QaIHhfRWPZ8TVRjijTn5Ci1iEabtTEWrt9CCo='; default-src 'self'; base-uri moz-extension://*;",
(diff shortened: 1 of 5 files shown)


src/manifest.json
```diff
@@ -30,7 +30,7 @@
     }
   },
 
-  "content_security_policy": "script-src 'self' 'sha256-MmZkN2QaIHhfRWPZ8TVRjijTn5Ci1iEabtTEWrt9CCo='; default-src 'self'; base-uri moz-extension://*;",
+  "content_security_policy": "script-src 'self' 'sha256-MmZkN2QaIHhfRWPZ8TVRjijTn5Ci1iEabtTEWrt9CCo='; default-src 'self'; base-uri moz-extension://*; object-src 'none'",
```
ksy36: I've noticed this has been added in m-c repository

miketaylr: yeah, looks like it was added here https://bugzilla.mozilla.org/show_bug.cgi?id=1499354

PR #87, opened 2019-10-07
Bug 1585380 - Add support for a control experiment branch

wisniewskit:
This adds a control experiment branch for the enhanced search feature, as well as having it emit the addon version to fennec telemetry, and only send extra telemetry when a user is in the experiment (or triggers the testing pref).

Note there is also a minor change injections.js to keep mach lint from complaining.

Also note that any necessary companion changes to the Fennec java files will be made in a separate patch on bug 1585380.

### DIFF
--- a/src/lib/ua_overrides.js
+++ b/src/lib/ua_overrides.js
@@ -54,6 +54,15 @@ class UAOverrides {
 
     const { blocks, matches, telemetryKey, uaTransformer } = override.config;
     const listener = details => {
+      // We set the "used" telemetry key if the user would have had the
+      // override applied, regardless of whether it is actually applied.
+      if (!details.frameId && override.shouldSendDetailedTelemetry) {
+        // For now, we only care about Telemetry on Fennec, where telemetry
+        // is sent in Java code (as part of the core ping). That code must
+        // be aware of each key we send, which we send as a SharedPreference.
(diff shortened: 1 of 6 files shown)


src/lib/ua_overrides.js
```diff
@@ -164,11 +178,20 @@ class UAOverrides {
...
 
-        // If there is a specific experiment running for
-        // this override, get its state.
+        // Note whether the user is actively in the override's experiment (if any).
+        override.experimentActive = false;
         const experiment = override.config.experiment;
-        override.experimentActive =
-          experiment && (await browser.experiments.isActive(experiment));
+        if (experiment) {
+          const branches = Array.isArray(experiment)
```
miketaylr: Can we comment on what's happening here? It took my brain too long to parse it.

wisniewskit: Yep.

src/manifest.json
```diff
@@ -2,7 +2,7 @@
   "manifest_version": 2,
   "name": "Web Compat",
   "description": "Urgent post-release fixes for web compatibility.",
-  "version": "6.1.0",
+  "version": "6.2.0",
```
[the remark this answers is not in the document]
wisniewskit: Good eye. I'll use 6.3.0.

PR #88, opened 2019-10-09
Bug 1587558 - Remove unneeded webkit scrollbar intervention for breitbart.com

miketaylr:
No longer needed: https://github.com/webcompat/web-bugs/issues/25156#issuecomment-540123555

r? @denschub


miketaylr:
Thanks!

PR #92, opened 2019-12-20
Bug 1598198 - Add User Agent modification for galaxy.store URLs

denschub:
~~**Do not merge**.~~

~~This works, but needs some more testing. I'll tackle this when I'm back at work after my vacation, but I wanted to dump this code onto GitHub, just in case.~~

Ready for review. r? @ksy36

### DIFF
--- a/src/data/ua_overrides.js
+++ b/src/data/ua_overrides.js
@@ -4,7 +4,7 @@
 
 "use strict";
 
-/* globals module, require */
+/* globals browser, module, require */
 
 // This is a hack for the tests.
 if (typeof getMatchPatternsForGoogleURL === "undefined") {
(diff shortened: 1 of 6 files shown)


src/manifest.json
```diff
@@ -43,16 +43,20 @@
...
   },
 
   "content_security_policy": "script-src 'self' 'sha256-MmZkN2QaIHhfRWPZ8TVRjijTn5Ci1iEabtTEWrt9CCo='; default-src 'self'; base-uri moz-extension://*; object-src 'none'",
 
-  "permissions": [
-    "webRequest",
-    "webRequestBlocking",
-    "<all_urls>"
-  ],
+  "permissions": ["webRequest", "webRequestBlocking", "<all_urls>"],
```
denschub: Prettier, apparently, changed its mind again.

src/data/ua_overrides.js
```diff
@@ -549,6 +549,38 @@ const AVAILABLE_UA_OVERRIDES = [
...
+    bug: "1598198",
+    config: {
+      matches: ["*://galaxy.store/*"],
+      uaTransformer: originalUA => {
+        if (!browser.systemManufacturer) {
+          return originalUA;
+        }
+
+        let manufacturer = browser.systemManufacturer.getManufacturer();
+        if (manufacturer && manufacturer.toLowerCase() == "samsung") {
```
miketaylr: Note for posterity: I asked our Samsung contact if it will always return `Samsung` and they said for all the devices they care about, yes. It's possible some really old Samsung phones on Android returned something different, but we don't need to care about them (per Samsung).

denschub: Ah, that's great, thanks for the info. Then this should be ready after a single safety-check by our Softvision friends. :) I'll prepare builds!

denschub:
Our contacts at Samsung have confirmed that the patch is working to their liking. I have confirmed that the new changes are not breaking desktop or Fenix. This is ready for review.

r? @ksy36 :)

ksy36:
Looks good to me :)

src/data/ua_overrides.js
```diff
@@ -549,6 +549,42 @@ const AVAILABLE_UA_OVERRIDES = [
...
+      ],
+      uaTransformer: originalUA => {
+        if (!browser.systemManufacturer) {
+          return originalUA;
+        }
+
+        let manufacturer = browser.systemManufacturer.getManufacturer();
+        if (manufacturer && manufacturer.toLowerCase() == "samsung") {
+          return originalUA.replace("Mobile;", "Mobile; Samsung;");
+        }
```
ksy36: nit pick, should it be `const manufacturer`? since the value is not reassigned 
also probably need to use triple equals `manufacturer.toLowerCase() === "samsung"`

denschub: Good points! Done. :)

PR #94, opened 2019-12-21
Bug 1605611 - Fix the time picker of Google Maps' transit section

Rob--W:
See https://bugzilla.mozilla.org/show_bug.cgi?id=1605611#c0 for details.

Fixes https://github.com/webcompat/web-bugs/issues/39281

### DIFF
--- a/src/data/injections.js
+++ b/src/data/injections.js
@@ -4,7 +4,12 @@
 
 "use strict";
 
-/* globals module */
+/* globals module, require */
+
+// This is a hack for the tests.
+if (typeof getMatchPatternsForGoogleURL === "undefined") {
+  var getMatchPatternsForGoogleURL = require("../lib/google");
+}
 
 /**
  * For detailed information on our policies, and a documention on this format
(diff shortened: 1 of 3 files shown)


Rob--W:
I verified that the patch works, by entering the `src/` directory and running the following:

```
web-ext run -t firefox-android --android-device=deviceidhere --firefox-apk=org.mozilla.fennec_aurora
```

This runs Firefox Nightly on my phone. Then I visited Google Maps and verified that the "Leave now" button can be clicked, and that the departure/arrival date can be set/cleared/changed as desired.

(to get the add-on to run properly, I had to take some more steps, because experiments are not available by default. If you want to know how to test this yourself, send me an e-mail)

miketaylr:
Thanks for the PR! Let me redirect review to @ksy36, she's taking point on interventions for the next release.

src/injections/js/bug1605611-maps.google.com-directions-time.js
```diff
@@ -0,0 +1,82 @@
...
+ */
+
+// Step 1.
+document.addEventListener("DOMContentLoaded", () => {
+  // In case the element appeared before the MutationObserver was activated.
+  for (let elem of document.querySelectorAll('.ml-directions-time[disabled]')) {
+    elem.disabled = false;
+  }
+  // Start watching for the insertion of the "Leave now" button.
+  let moOptions = {
```
ksy36: If `moOptions` variable is not reassigned later, could you please use `const` instead of `let`? 

There are multiple occurrences of `let` in this file, they should all probably be `const`, except `let restore = false;`  (since it's reassigned later to be `restore = true;` )

src/data/injections.js
```diff
@@ -4,7 +4,12 @@
 
 "use strict";
 
-/* globals module */
+/* globals module, require */
+
+// This is a hack for the tests.
+if (typeof getMatchPatternsForGoogleURL === "undefined") {
+  var getMatchPatternsForGoogleURL = require("../lib/google");
```
ksy36: nit: could you please use `const` here as well?

Rob--W: No, this is intended to be a global variable declaration. Using `const` instead of `var` would limit the import to the block scope, which will probably cause the test to fail.

This use of `var` (and the comment before it) is copied from src/data/ua_overrides.js .

ksy36: Ahh right, my bad

PR #95, opened 2020-01-09
Bug 1605611 - `npm run prettier`-run to prevent styleguide backouts on mozilla-central.

denschub:
This is a style-only PR to prevent build-errors and backouts when trying to land this stuff.

r? @ksy36


denschub:
Ignore. I'll do something else and submit a PR soon.

PR #105, opened 2020-03-03
Bug 1614068 - Apply Picture-in-Picture toggle policy to laracasts.com sites.

mikeconley:
This positions the toggle at one quarter from the top of the video, to avoid
the "next video" carousel button on laracasts.com.


denschub:
Merged as https://github.com/mozilla/webcompat-addon/commit/46886affdf3350c4d5a31df13fc8fb744b182a7f.

There were spaces in the empty line, which prettier doesn't like, but I had prettier fix that. But because of that change, GitHub doesn't know I actually merged that commit. Hint: [We have a pre-push hook](https://github.com/mozilla/webcompat-addon#run-tests-and-codestyle-checks-automatically-before-pushing) to run prettier automatically!

Thanks for your PR. :)

PR #106, opened 2020-03-05
Bug 1620218 - Upgrade Prettier to 1.18.2.

denschub:
This is a preparation so we're ready whenever the changes happen on `mozilla-central`. Do not merge this PR before [bug 1620218](https://bugzilla.mozilla.org/show_bug.cgi?id=1620218) is landed.


denschub:
This landed upstream, and because this is a change that's already upstream, I'll merge this myself without reviews. 👿

PR #111, opened 2020-03-10
Bug 1621065 - Write UA override for bracketchallenge.ncaa.com

miketaylr:
If we don't have "Chrome" in the UA string, we're detected as
an older version of Android and our users get a scary modal.

I've tested locally on Fennec and the Sample browser from AC.

r? @ksy36


miketaylr:
whoops, copypasta test failure!

miketaylr:
https://github.com/webcompat/web-bugs/issues/49886

miketaylr:
thanks!

PR #116, opened 2020-03-13
Bug 1620556 - Upgrade Prettier to 1.19.x

denschub:
This will stay a draft PR until [this Phabricator revisoin](https://phabricator.services.mozilla.com/D66128) is landed. It's likely this will need a quick rebase with the upcoming changes to this addon, but it shouldn't be too bad (the second commit is automated via `npm run prettier` anyway. :))


denschub:
Changeset landed upstream, r=me because of that.

PR #119, opened 2020-03-19
Bug 1622059 - Add UA override for sites using img.weblogssl.com script

miketaylr:
I've verified this fixes the bug as reported -- comments can be viewed, and toggled.

(The version isn't used at all, but if it's not there the page throws and breaks)

### DIFF
--- a/src/data/ua_overrides.js
+++ b/src/data/ua_overrides.js
@@ -626,6 +626,33 @@ const AVAILABLE_UA_OVERRIDES = [
       },
     },
   },
+  {
+    /*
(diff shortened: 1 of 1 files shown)


ksy36:
Looks good to me. Maybe we could add two more domains here? Looks like the same script is used there

www.xataka.com
https://github.com/webcompat/web-bugs/issues/48650

www.xatakandroid.com
https://github.com/webcompat/web-bugs/issues/48787

miketaylr:
Ah, right on. Thanks.

miketaylr:
> Looks good to me.

This sounds like permission to self-merge. :)

PR #122, opened 2020-03-23
Bug 1622062 - Rewrite preventDefault:true for $.detectSwipe

miketaylr:
r? @ksy36 

Tested and working on a custom Fennec build.

(Note: going to file a follow-up bug, because it seems like some of this could be abstracted)

### DIFF
--- a/src/lib/custom_functions.js
+++ b/src/lib/custom_functions.js
@@ -36,6 +36,25 @@ const replaceStringInRequest = (
 };
 
 const CUSTOM_FUNCTIONS = {
+  detectSwipeFix: injection => {
+    const { urls, types } = injection.data;
+    const listener = (injection.data.listener = ({ requestId }) => {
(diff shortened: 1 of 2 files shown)


src/lib/custom_functions.js
```diff
@@ -36,6 +36,25 @@ const replaceStringInRequest = (
...
+      );
+      return {};
+    });
+    browser.webRequest.onBeforeRequest.addListener(listener, { urls, types }, [
+      "blocking",
+    ]);
+  },
+  detectSwipeFixDisable: injection => {
+    const { listener } = injection.data;
+    browser.webRequest.onHeadersReceived.removeListener(listener);
```
ksy36: should the listener be removed from `onBeforeRequest`?

miketaylr: whoops!

PR #124, opened 2020-03-25
Bug 1621062 - Bump version to 9.0.0

miketaylr:
r? @denschub 

(my prettier-vscode installation insisted on formatting the JSON file, seems ok?)

### DIFF
--- a/src/manifest.json
+++ b/src/manifest.json
@@ -2,7 +2,7 @@
   "manifest_version": 2,
   "name": "Web Compat",
   "description": "Urgent post-release fixes for web compatibility.",
-  "version": "8.0.0",
+  "version": "9.0.0",
 
   "applications": {
     "gecko": {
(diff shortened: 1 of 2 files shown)


src/manifest.json
```diff
@@ -89,4 +128,4 @@
       "run.js"
     ]
   }
-}
+}
```
denschub: Did prettier remove the newline at the end? That's... odd.

miketaylr: (my VS-code might have some funky settings....)

miketaylr:
Hm, why is abde5da in this PR????

denschub:
r+ for the one commit updating the manifest. I'll do the merge and cherry-pick only the right commit, so you don't have to fight with git. One second. :)

miketaylr:
thanks @denschub!

PR #125, opened 2020-03-25
Also run prettier on .json files

denschub:
mozilla-central is not (yet?) doing that, but since prettier can work on .json files and is doing a fairly good job of keeping them readable, let's remove one additional point of ambiguity and have our embedded prettier do the job for us.

r? @miketaylr


miketaylr:
thank you!

PR #128, opened 2020-04-17
Bug 1625222 - Interventions and overrides for v10.0.0 

ksy36:
r? @miketaylr


src/data/injections.js
```diff
@@ -467,6 +467,20 @@ const AVAILABLE_INJECTIONS = [
       ],
     },
   },
+  {
+    id: "bug1629926",
+    platform: "desktop",
+    domain: "hangouts mic",
+    bug: "1629926",
+    contentScripts: {
+      matches: ["https://mail.google.com/*"],
```
miketaylr, line 433: Does the issue only happen in Gmail, or does this need to apply on hangouts.google.com? (I'm not sure...)

ksy36: Yeah, the problem is only when calling from within the gmail interface. Calling works when calling on hangouts.google.com directly

miketaylr: Sweet, thanks!

src/data/injections.js
```diff
@@ -438,6 +395,49 @@ const AVAILABLE_INJECTIONS = [
...
+        {
+          file: "injections/css/bug1630280-dominos.ch-hide-overlay.css",
+        },
+      ],
+    },
+  },
+  {
+    id: "bug1629926",
+    platform: "desktop",
+    domain: "hangouts mic",
```
wisniewskit, line 430: Do we want to be a little more clear on the fix here, and say "GMail Hangouts microphone fix", or just say `mail.google.com` instead to be accurate to the domain?

src/injections/js/bug1629926-mail.google.com-hangouts-mic-fix.js
```diff
@@ -0,0 +1,59 @@
...
+  let n = 0;
+  createObserver((records, observer) => {
+    const _element = getElementFn();
+    if (_element) {
+      onFound(_element);
+      observer.disconnect();
+      return;
+    }
+
+    if (n > limit) {
```
wisniewskit, line 47: Do we really want a limit here? Will users who leave GMail open and make many calls end up with the microphone not working?

src/data/injections.js
ksy36: Yeah, good point, thanks!

src/injections/js/bug1629926-mail.google.com-hangouts-mic-fix.js
ksy36: So this observer is waiting for the iframe element to be in the dom on initial load (once it's located, "allow" attribute is added). The limit is just to be safe, so we don't continue to poll in case the iframe with id "gth-talk-plugin-frame-id" is never found. 

I've tried making multiple calls and since the targeted iframe is never removed/modified, it works as expected

wisniewskit: Ah, I see. Maybe it would be better to just add a Mutation Observer right away which listens for attribute changes, and when it sees an `id` change to `gth-talk-plugin-frame-id` for an `iframe`, it adds our `allow` value? If that works, we should be able to avoid polling and just leave the MutationObserver active just in case (or disconnect it then, if the iframe seems to never be re-created by the page).

ksy36:
After talking to Tom decided to change the Gmail intervention to what he initially suggested in https://bugzilla.mozilla.org/show_bug.cgi?id=1503694#c37 (with waiting for src to be set, instead of allow)

ksy36:
@wisniewskit @miketaylr  could you take another look?

src/injections/js/bug1629926-mail.google.com-hangouts-mic-fix.js
```diff
@@ -0,0 +1,40 @@
...
+
+/* globals exportFunction */
+
+console.info(
+  "hangouts iframe allow property was changed for compatibility reasons. See https://bugzilla.mozilla.org/show_bug.cgi?id=1503694 for details."
+);
+
+const isHangoutsSrc = src => {
+  const url = new URL(src);
+  return url.host === "hangouts.google.com" && src.includes("blank");
```
wisniewskit, line 22: If we can use something like `url.path.includes("/blank")`, that might be better, but it's probably a minor nit.

PR #137, opened 2020-05-15
Bug 1546051 - Hide PiP on Netflix previews

miketaylr:
OK, so Netflix has a lot of domains. I've tested all of these (via curl and commandline fetch hax), cobbled from a few sources. I'm slightly concerned that we're increasing memory usage by adding 200 new PiP policy overrides for all our users, when they likely only need this for a single Netflix TLD.

r? @denschub 

Also cc @adamopenweb, just to see if he thinks we should trim the list of Netflix domains.

### DIFF
--- a/src/data/picture_in_picture_overrides.js
+++ b/src/data/picture_in_picture_overrides.js
@@ -28,6 +28,9 @@ let AVAILABLE_PIP_OVERRIDES;
     // Laracasts
     "https://*.laracasts.com/*": TOGGLE_POLICIES.ONE_QUARTER,
 
+    // Netflix
+    "https://*.netflix.com/browse": TOGGLE_POLICIES.HIDDEN,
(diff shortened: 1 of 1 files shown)


miketaylr:
Also cc @mikeconley to maybe give some input on the memory usage aspects of this. Another approach is to just fix this for a subset of TLDs, where Firefox usage is high.

miketaylr:
However, @adamopenweb just noticed that even in Canada netflix.com/browse is used. So maybe I'm entirely overthinking this.

miketaylr:
> However, @adamopenweb just noticed that even in Canada netflix.com/browse is used. So maybe I'm entirely overthinking this.

Yeah, seems like that's true for Canada, Germany, and the UK. A safe bet that we can just use .com/browse.

PR #140, opened 2020-06-22
Bug 1637672 - A couple of changes for version 12.

denschub:
This adds all interventions currently marked for v12 - with one exception: the twitter tracking protection intervention. That one may have to wait, because I can't test the broken state without further information (see https://bugzilla.mozilla.org/show_bug.cgi?id=1641998#c1)

r? @ksy36


src/data/injections.js
```diff
@@ -411,6 +411,21 @@ const AVAILABLE_INJECTIONS = [
       ],
     },
   },
+  {
+    id: "bug1644830",
+    platform: "desktop",
+    domain: "usps.com",
+    bug: "1644830",
+    contentScripts: {
+      matches: ["https://*.usps.com/*"],
```
denschub, line 420: Note: I'm adding the wildcard here, as this probably is not limited to a specific subdomain. The registration is at `reg.usps.com`, but `missingmail.usps.com` also has affected checkboxes - so it's probably a save bet that there are other forms affected.

PR #141, opened 2020-06-23
Bug 1641998 - Overwride IndexedDB APIs for embedded Twitter videos.

denschub:
Follow-up to yesterday, as I didn't get required information in time.

r? @ksy36


denschub:
The linter here was happy, who knows what the m-c linter says. It's magical.

PR #144, opened 2020-06-30
Bug 1649325. Adds cloudreadykb.neverware.com to salesforce issue

karlcow:
This will fix the issue with the donotshowunsupported message cloudreadykb.neverware.com.

### DIFF
--- a/src/data/injections.js
+++ b/src/data/injections.js
@@ -132,6 +132,7 @@ const AVAILABLE_INJECTIONS = [
           "https://help.doordash.com/*",
           "https://community.snowflake.com/*",
           "https://tivoidp.tivo.com/*",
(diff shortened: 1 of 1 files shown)


karlcow:
@denschub the taskcluster is not starting, is it expected. Pull request from my fork.

denschub:
> the taskcluster is not starting, is it expected

*Technically* yes, because that's what the template does by default. That kinda makes sense for some projects that use BORS, but I don't see us having a need for this level of complexity. [I changed that](https://github.com/mozilla-extensions/webcompat-addon/commit/ebe46046170bc2ebdddbc9d989570631ab05d906), and I'll gently poke your PR so that taskcluster builds.

PR #146, opened 2020-07-08
Bug 1651292. Adds UA override to www.jp.square-enix.com

karlcow:
This will give the possibility for the page to load on https://www.jp.square-enix.com/music/sem/page/FF7R/ost/

### DIFF
--- a/src/data/ua_overrides.js
+++ b/src/data/ua_overrides.js
@@ -620,6 +620,26 @@ const AVAILABLE_UA_OVERRIDES = [
       },
     },
   },
+  {
+    /*
+     * Bug 1651292 - UA override for www.jp.square-enix.com
+     * Webcompat issue #53018 - https://webcompat.com/issues/53018
+     *
+     * Unless the UA string contains "Chrome 66+", a section of
+     * www.jp.square-enix.com will show a never ending LOADING
+     * page.
+     */
+    id: "bug1651292",
+    platform: "android",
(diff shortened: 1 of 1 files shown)


src/data/ua_overrides.js
```diff
@@ -620,6 +620,26 @@ const AVAILABLE_UA_OVERRIDES = [
...
+     * page.
+     */
+    id: "bug1651292",
+    platform: "android",
+    domain: "www.jp.square-enix.com",
+    bug: "1651292",
+    config: {
+      matches: ["*://www.jp.square-enix.com/music/sem/page/FF7R/ost/*"],
+      uaTransformer: originalUA => {
+        return originalUA + " Chrome 66";
```
miketaylr, line 639: drive by comment: Chrome right now is at 83, it might be safer to choose something closer to that (in case they decide in 1 year to change minimal support to 80+ or something).

ksy36: yeah, this is probably a valid point, lets make it closer to 83 just to be safe :)

karlcow:
```js
var e = navigator.userAgent
undefined
e = e + ' Chrome/83'
"Mozilla/5.0 (Android 10; Mobile; rv:80.0) Gecko/80.0 Firefox/80.0 Chrome/83"
e.indexOf('Android') > 0
true
e.match(/Chrome\/([\d]+)/).slice(1) >= 66
true
```

PR #148, opened 2020-07-10
Bug 1651917 - Adds CSS intervention for teletrader.com

karlcow:
Fix the body transform origin position for teletrader.com
See https://bugzilla.mozilla.org/show_bug.cgi?id=1651917
and https://github.com/webcompat/web-bugs/issues/55217

### DIFF
--- a/src/data/injections.js
+++ b/src/data/injections.js
@@ -457,6 +457,21 @@ const AVAILABLE_INJECTIONS = [
       ],
     },
   },
+  {
+    id: "bug1651917",
+    platform: "android",
+    domain: "teletrader.com",
+    bug: "1651917",
+    contentScripts: {
+      matches: ["*://*.teletrader.com/*"],
+      css: [
+        {
+          file:
+            "injections/css/bug1651917-teletrader.com.body-transform-origin.css",
+        },
+      ],
+    },
+  },
 ];
 
 module.exports = AVAILABLE_INJECTIONS;
--- a/src/injections/css/bug1651917-teletrader.com.body-transform-origin.css
+++ b/src/injections/css/bug1651917-teletrader.com.body-transform-origin.css
@@ -1,0 +1,14 @@
+/**
+ * teletrader.com - content is shifted down and right
+ * Bug #1651917 - https://bugzilla.mozilla.org/show_bug.cgi?id=1651917
+ * WebCompat issue #55217 - https://webcompat.com/issues/55217
+ *
+ * The content is shifted down and right, because they use webkit prefixes
+ * for scaling and redefining the origin. Firefox doesn't support
+ *     -webkit-transform-origin-x/y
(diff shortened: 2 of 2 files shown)


denschub:
Just a quick drive-by. I mentioned this on Slack, but it got buried in the noise: Do you mind using the "Bug nnn - Foo." commit format instead of "Bug nnn. Foo."? Small detail, but *usually* we use the Gecko-commit format here, and you'll have to use the dash-format for landing in m-c anyway, so it's probably best to stick with that. :)

miketaylr:
> Just a quick drive-by. I mentioned this on Slack, but it got buried in the noise: Do you mind using the "Bug nnn - Foo." commit format instead of "Bug nnn. Foo."? Small detail, but _usually_ we use the Gecko-commit format here, and you'll have to use the dash-format for landing in m-c anyway, so it's probably best to stick with that. :)

I think we can (and should!) decide on a commit format for this project, and `Bug NNN - Message` is a great one, but I don't think using a dot or a colon actually matters for landing. 

(I was trying to find commit message guidelines and the only thing I could find was in Step 3 of https://developer.mozilla.org/en-US/docs/Mozilla/Developer_guide/Introduction):

Here's one with no separator  between the bug number and the message:
https://hg.mozilla.org/mozilla-central/rev/e3163482cc62d0c7aed592c6edcd733cf9475330

Here's one using a colon:
https://hg.mozilla.org/mozilla-central/rev/9d3a0ea2cf657a837ad38f43e3cad5eb498fad8f

Here's one with a dot:
https://hg.mozilla.org/mozilla-central/rev/40792c588f9b27cd0012543b2d0a76aa07e932ef

That said the vast majority use `Bug NN - Message`.

(If this were a big problem, I imagine it would be handled at the Phabricator level where they rewrite your commit message anyways, after it gets r+)

denschub:
Interesting - https://firefox-source-docs.mozilla.org/contributing/contribution_quickref.html#to-write-a-patch suggests the message format. That being said, I have no interest in bikeshedding over this, I just figured I can mention it, since we mostly stick to the dash here (with the exception of the template commits that have been merged into) - and I only mentioned this in a discussion about the Google doc recently shared, where I think the message got lost in traffic. I'm fine with ignoring this. (Also, this may not be the best place to have this discussion :/)

karlcow:
ok changed the commit message for this one. 


fwiw

```
~/code/webcompat-addon-karlcow % git log --oneline | egrep -i "^....... Bug" | egrep -iv " - "
d29478e Bug 1651292. Adds UA override to www.jp.square-enix.com
4770111 Bug 1649325. Adds cloudreadykb.neverware.com to salesforce issue
4d27137 Bug 1454202: Follow-up: Fix straggler callback-based API caller. r=me,test-only DONTBUILD
f5e4fee Bug 1291416. Mark the webcompat system add-on as e10s compatible as it is just a stub. r=mconley
```

PR #152, opened 2020-07-13
Bug 1652025 - Remove Intervention for "files with no content-type"

karlcow:
see https://bugzilla.mozilla.org/show_bug.cgi?id=1652025

I was not sure if I should keep or remove the code inside the injection-spec.js

### DIFF
--- a/src/data/injections.js
+++ b/src/data/injections.js
@@ -160,23 +160,6 @@ const AVAILABLE_INJECTIONS = [
     customFunc: "pdk5fix",
   },
   {
-    id: "bug1577870",
-    platform: "desktop",
-    domain: "Download prompt for files with no content-type",
-    bug: "1577870",
-    data: {
-      urls: [
-        "https://ads-us.rd.linksynergy.com/as.php*",
-        "https://www.office.com/logout?sid*",
-      ],
-      contentType: {
-        name: "content-type",
-        value: "text/html; charset=utf-8",
-      },
-    },
-    customFunc: "noSniffFix",
-  },
-  {
     id: "bug1561371",
     platform: "android",
     domain: "mail.google.com",


src/data/injections.js
```diff
@@ -159,23 +159,6 @@ const AVAILABLE_INJECTIONS = [
...
-      urls: [
-        "https://ads-us.rd.linksynergy.com/as.php*",
-        "https://www.office.com/logout?sid*",
-      ],
-      contentType: {
-        name: "content-type",
-        value: "text/html; charset=utf-8",
-      },
-    },
-    customFunc: "noSniffFix",
```
ksy36, line 177: We also need to remove the custom function (noSniffFix and noSniffFixDisable):
https://github.com/mozilla-extensions/webcompat-addon/blob/master/src/lib/custom_functions.js#L58-L74

karlcow: @ksy36  Thanks.

https://github.com/mozilla-extensions/webcompat-addon/blob/master/src/lib/custom_functions.js#L58-L74

What about 

https://github.com/mozilla-extensions/webcompat-addon/blob/72e35293a0e4f4a0c34af908455fedb3ea83ffc9/spec/injections.spec.js#L39-L54

and 

https://github.com/mozilla-extensions/webcompat-addon/blob/72e35293a0e4f4a0c34af908455fedb3ea83ffc9/spec/injections.spec.js#L152-L159

Just to understand what is to be removed or not.

ksy36: Yeah, good point, this test and one below it can be removed

ksy36:
@karlcow I realized that we may want to keep this intervention, see my comment https://bugzilla.mozilla.org/show_bug.cgi?id=1652025#c2
I'll try to test the office365 intervention once I find the credentials

miketaylr:
> see my comment https://bugzilla.mozilla.org/show_bug.cgi?id=1652025#c2

I asked Oana to re-test, so let's land this in the 14 cycle.

ksy36:
> I asked Oana to re-test, so let's land this in the 14 cycle.

Thanks! I've tried to login with softvision testing account to office365, but there is two factor authorization asking to verify my account via text message :/

ksy36:
Looks like this intervention is still needed for one of the sites listed (https://bugzilla.mozilla.org/show_bug.cgi?id=1652025#c4)

I've taken care of this in v14.0.0 https://github.com/mozilla-extensions/webcompat-addon/pull/160/files#diff-c4d06abd5f522062438517d128b3f90eR125, so I think we can close this PR @karlcow

karlcow:
Thanks a lot @ksy36 🌞

PR #154, opened 2020-07-17
Bug 1652396 - Remove UA Override for cineflix.com.br

karlcow:
This fixes https://bugzilla.mozilla.org/show_bug.cgi?id=1652396


miketaylr:
Let's go ahead and merge even though decision task is still running. 🤷

PR #166, opened 2020-09-09
Add ability to disable picture-in-picture interventions. Fixes #165

### LINKED ISSUE #165: Import changes allowing easy disabling of PIP interventions.
That is, this patch should be imported: https://phabricator.services.mozilla.com/D87182

wisniewskit:


denschub (changes requested):
This is missing the manifest-version bump to `14.1.0`, and it's probably wise to include that. :)

PR #169, opened 2020-09-09
Reduce some logspam caused by the shim support code. Fixes #167

### LINKED ISSUE #167: Reduce some logspam caused by the shim support code
See https://bugzilla.mozilla.org/show_bug.cgi?id=1660108 for context.

Basically, sometimes trying to access `topLevelUrl` fails here: https://searchfox.org/mozilla-central/rev/6cc48251bb97600fdf11a5b4c5f621bfc8606d55/browser/extensions/webcompat/experiment-apis/trackingProtection.js#41

It seems the easiest fix for now is to just wrap the access to `topLevelUrl` in a try/catch block, and since `new URL` could similarly fail some day, it's probably good defensive programming to wrap that, too.

wisniewskit:


denschub (approved):

miketaylr (approved):
lgtm, thanks

PR #175, opened 2020-09-09
Create an intervention for reallygoodemails.com. Fixes #174

### LINKED ISSUE #174: Create an intervention for reallygoodemails.com
As per https://bugzilla.mozilla.org/show_bug.cgi?id=1654906 and https://github.com/webcompat/web-bugs/issues/55001, the site expects iframes to never be in an uninitialized readyState, and just writes to their innerHTML even while they are still in that state in Firefox (rather than waiting for the iframe's load event to fire first). The result is that their HTML email preview layer is often blank on first page-load.

An intervention which simply listens for their attempts to set innerHTML on an iframe, and defers until the iframe has loaded first, seems to be working fine, so let's try that.

wisniewskit:


denschub:
Ah interesting, this is a fix for whatwg/html#4965 and https://bugzilla.mozilla.org/show_bug.cgi?id=728151, which I've now seen at least twice.

denschub (approved):
This will break in funny ways if the site ever decides to use an actual iFrame with a src, but this seems like a good workaround for this specific case.

lgtm, r+!

Issue #180, opened 2020-09-11
Import changes from m-c enabling keyboard hotkey controls for picture-in-picture interventions.


wisniewskit:
That is, import the addon-specific changes from https://bugzilla.mozilla.org/show_bug.cgi?id=1662576, once that lands.

PR #183, opened 2020-09-14
Fix minor mis-merge from the fix to #180

wisniewskit:


miketaylr (approved):
You're just trying to make me look bad as a reviewer, admit it.

PR #187, opened 2020-10-07
Bug 1653296 - Don't override the UA on mobile if Request Desktop Site is enabled.

denschub:
As discussed in https://bugzilla.mozilla.org/show_bug.cgi?id=1653296, this is probably the easiest way. The UA will get manipulated for the Desktop Mode anyway, so there isn't really a point to us doing overrides.

I honestly don't know how to test this, since our unit tests do not actually fire the handler. The change can be verified using the testbed, but that's not ideal. We probably should move more tests into the browser runtime at some point. :/

r? @ksy36


wisniewskit (approved):
Drive-by review for fun :)

PR #192, opened 2020-10-09
Bug 1670241 - Re-ship console notifications.

denschub:
This is blocked by https://bugzilla.mozilla.org/show_bug.cgi?id=1670230 for now.

r=me, as this is just something that previously landed.


denschub:
I'll CLOSE WONTFIX this for now. There is no urgent need to solve this, the original patch is old, and we have some planned refactoring hopefully coming up anyway.

PR #196, opened 2020-11-02
Fixes 1663978 - ship v17.0.0 site interventions

karlcow:


src/data/ua_overrides.js
```diff
@@ -677,6 +677,42 @@ const AVAILABLE_UA_OVERRIDES = [
...
+    platform: "all",
+    domain: "nha.chotot.com",
+    bug: "1671150",
+    config: {
+      matches: ["*://nha.chotot.com/*"],
+      uaTransformer: originalUA => {
+        return originalUA.replace("Firefox", "WebcompatFox");
+      },
+    },
+  },
```
denschub, line 715: We don't need this anymore, the site luckily fixed it! 🥳

denschub (changes requested):
Looking good, but the chotot.com-override appears to be no longer needed, so let's remove that one :) (Also needs a rebase, because I fixed the version number!)

src/data/ua_overrides.js
karlcow: fixed

denschub (approved):

PR #197, opened 2020-11-19
Bug 1663980 - v18.0.0 of webcompat system addon

ksy36:
r? @denschub


package-lock.json
```diff
@@ -5999,8 +5998,7 @@
           "version": "5.1.2",
           "resolved": "https://registry.npmjs.org/safe-buffer/-/safe-buffer-5.1.2.tgz",
           "integrity": "sha512-Gd2UZBJDkXlY7GbJxfsE8/nvKkUEU1G38c1siN6QP6a9PT9MmHB8GnpscSmMJSoF8LOIrt8ud/wPtojys4G6+g==",
-          "dev": true,
-          "optional": true
+          "dev": true
         },
         "string_decoder": {
```
denschub, line 6003: Huh, is this something that always happens to you? I don't think it should be in this patch, unless that can't be avoided. :)

denschub (commented):
Looking good, one comment though!

package-lock.json
ksy36: Hmm, no I think it's the first time :) I've removed it

ksy36:
Thanks!

denschub (approved):

PR #200, opened 2020-12-14
Bug 1682238 - gamearter.com: Override macOS version number in UA string from 11.0 to 10.16.

denschub:
r? @ksy36


denschub:
Confirmed to be working on the M1 Macs by :tbabos, https://bugzilla.mozilla.org/show_bug.cgi?id=1680516#c18

ksy36 (approved):

ksy36:
Oops, missed that the version should be updated

PR #204, opened 2021-01-26
Bug 1687944 - log a message to the web console when users opt in to a…

wisniewskit:
…llowing trackers via shims


denschub (commented):
This looks good to me - however, linting fails, see https://firefox-ci-tc.services.mozilla.com/tasks/EFDcdSvlQQGSQTAsKYD05g/runs/0/logs

Could you look into that?

denschub (approved):
r+ assuming CI is happy.

PR #210, opened 2021-03-15
Bug 1693824 - Ship v21.0.0 of webcompat system addon

karlcow:
Cleaning and decoration staff 

r? @denschub


denschub:
Looking good! And thanks for the dependency updates as well.

r+ :shipit:

denschub (approved):

PR #215, opened 2021-04-13
Bug 1702011 - Ship v22.0.0 of webcompat system addon

ksy36:
r? @wisniewskit @denschub


src/manifest.json
```diff
@@ -2,7 +2,7 @@
   "manifest_version": 2,
   "name": "@EXTENSION_NAME@",
   "description": "Urgent post-release fixes for web compatibility.",
-  "version": "20.1.0",
+  "version": "22.0.0",
```
ksy36, line 6: As the version wasn't updated in https://github.com/mozilla-extensions/webcompat-addon/pull/210, I suppose we skip the v. 21? Or we can make this v. 21 :)

denschub: Urgh everything is confusing. :) In-tree, the version number is actually currently at v21.3.0 - https://searchfox.org/mozilla-central/rev/759872688df15a5d6ab305ffe39d90450590bfec/browser/extensions/webcompat/manifest.json#5

Somehow, we missed uplifting a couple of patches. This isn't great, and it may even be my fault. :( I'll have a look at the missing patches tomorrow and then uplift everything that's missing. However, that means this PR will need rebasing.

After that, if you want to ship that in 89, I think we still should make it version 22, and then update the tracking OKR so that 23 is for Firefox 90 (and that it's not your turn again). But let's worry about re-syncing the repos first.

wisniewskit: It's worth noting that I will be backporting a couple of patches as soon as they both land in m-c:
- https://bugzilla.mozilla.org/show_bug.cgi?id=1705111
- https://phabricator.services.mozilla.com/D104697

Those are currently landing in m-c as version 21.4.0.

wisniewskit (approved):

PR #219, opened 2021-05-18
Bug 1711082 - Cannot trigger search bar on AliExpress mobile page

wisniewskit:
…ksenia

### DIFF
--- a/src/injections/js/bug1711082-m.aliexpress.com-undisable-search.js
+++ b/src/injections/js/bug1711082-m.aliexpress.com-undisable-search.js
@@ -1,0 +1,34 @@
+"use strict";
+
+/**
+ * Bug 1711082 - Cannot trigger search bar on AliExpress mobile page
+ *
+ * This patch ensures that the search input never has the [disabled]
+ * attribute, so that users may tap/click on it to search.
+ *
+ * See https://bugzilla.mozilla.org/show_bug.cgi?id=1711082 for details.
+ */
+
+const SELECTOR = `[data-spm="header"] input[disabled]`;
+
+function check(target) {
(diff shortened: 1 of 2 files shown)


src/injections/js/bug1711082-m.aliexpress.com-undisable-search.js
```diff
@@ -0,0 +1,34 @@
...
+      check(target);
+    } else {
+      addedNodes?.forEach(node => {
+        if (!check(node)) {
+          node.querySelector?.(SELECTOR)?.removeAttribute("disabled");
+        }
+      });
+    }
+  }
+}).observe(document, { attributes: true, childList: true, subtree: true });
```
ksy36, line 34: I think we might need to add `.disconnect()` to stop observing once target element was found? They are adding new elements to the page once you scroll, so probably a good idea to stop watching for changes

ksy36 (changes requested):
Thank for looking into this :) Looks good to me, except one small change is needed I think

src/injections/js/bug1711082-m.aliexpress.com-undisable-search.js
wisniewskit: Sure, i can do that, but I'm just not 100% sure if they ever add the [disabled] back to it, since they're using a react-style system to render it, and I see that they're messing with the element as you search. I'm not seeing a huge amount of traffic going to the observer as I scroll, so maybe it's worth just keeping it active just in case?

ksy36: Yeah, good point, I'm ok with not adding disconnect then

ksy36 (approved):

PR #228, opened 2021-07-09
Bug 1714611 - Ship v24.0.0 of webcompat system addon

denschub:
The usual fun, all in one PR, including a commit that updates the build deps to include security fixes.

This patch set does not include a version number bump - as usual, I bump the version number while landing in m-c and a-c, and only push the bump-commit to this repo once it's landed. This gives me the chance to publish eventual follow-up/fix-commits without having a messy history or burning a dot-release. :)

r? @wisniewskit

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -10,75 +10,51 @@
       "license": "MPL-2.0",
       "devDependencies": {
         "jake": "^10.4.3",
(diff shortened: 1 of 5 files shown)


wisniewskit (approved):
LGTM

wisniewskit (approved):

PR #229, opened 2021-08-02
Bug 1720138 - Ship v25.0.0 of webcompat system addon

karlcow:
These are the modifications of the webcompat system addon v25.0.0 for Firefox 92.


denschub (approved):
Looking good, thank you! :)

PR #232, opened 2021-08-30
Bug 1720139 - Ship v26.0.0 of webcompat system addon 

ksy36:
r? @wisniewskit


src/injections/js/bug1724868-news.yahoo.co.jp-ua-override.js
```diff
@@ -0,0 +1,25 @@
...
+ * content script allows playing the videos.
+ */
+
+/* globals exportFunction */
+
+console.info(
+  "The user agent has been overridden for compatibility reasons. See https://webcompat.com/issues/82605 for details."
+);
+
+Object.defineProperty(window.navigator.wrappedJSObject, "userAgent", {
```
wisniewskit, line 19: I don't mind this being here, but isn't it better as a UA override `in ua_overrides.js` rather than as a JS intervention? (That way it would appear in the about:compat UA override section and not require this extra file/etc).

ksy36: So yeah, I wasn't able to make the unsupported message go away with having this override in the `ua_overrides.js`.
The video loads from an iframe, and there are scripts in the frame that is performing UA detection. They seem to check the UA in two different places, and if either of these checks detects Firefox, they show the unsupported message.

While having the overrides in `ua_overrides.js` (overriding the header), it didn't seem to work on one of the scripts and it was still able to get the actual UA. So perhaps the fact that the override is in the content script, instead of the header makes it work.

wisniewskit: Interesting, we should probably investigate that at some point. But for now, r+.

wisniewskit (approved):

src/injections/js/bug1724868-news.yahoo.co.jp-ua-override.js
denschub: Overriding the header sadly does not affect what the `navigator` properties return. We could extend the logic of ua_overrides to always inject a JS override as well, but luckily, in 90% of the cases, spoofing the header is enough.

PR #234, opened 2021-09-02
No bug - Add bare-minimum specs for Shims.

denschub:
This is not meant to replace the in-product test or our work-in-progress test suite. but it provides a first stage of defense against common errors. The tests added here catch:

- Syntax errors in the definition file
- Missing or non-unique shim IDs
- Missing or invalid platform targets
- Missing bug IDs
- Missing or invalid URL match patterns

---

@wisniewskit and I briefly chatted about this before, but Tom didn't find the time so far, and I needed a bit of productivity right now. :)

r? @wisniewskit


wisniewskit (approved):
Thanks for doing this!

PR #242, opened 2021-09-28
V27dev

wisniewskit:


src/data/ua_overrides.js
```diff
@@ -47,7 +47,7 @@ const AVAILABLE_UA_OVERRIDES = [
     domain: "att.tv",
```
denschub, line 50: Nit: This isn't needed anywhere but the `about:compat` listing, but for reference, we should probably change the domain here as well. Feel free to pick one (but `directtv.com` is probably more meaningful?)

src/shims/stackblitz.js
```diff
@@ -0,0 +1,21 @@
...
+ *
+ * StackBlitz does not function correctly with Total Cookie
+ * Protection because it does not request storage access
+ * permissions before trying to register service workers.
+ * Until this is addressed, service workers are disabled
+ * for StackBlitz to mitigate the breakage.
+ */
+
+console.info(
+  `Service Workers are disabled by Firefox on this page due to https://bugzilla.mozilla.org/show_bug.cgi?id=1668408 for details.`
```
denschub, line 18: Nit: This doesn't need to be a template literal.

denschub (approved):
r+, two small nitpicks.

PR #247, opened 2021-11-10
Additional Interventions for Version 29

denschub:
I will build a couple of additional interventions for Karl's v29 cycle, mainly old stuff and things I stumble upon while diagnosis. The goal here is to manage our backlog a bit.

I'll convert this from a draft to a ready-to-review PR when it's close to landing time. I'll also keep this rebased if other things land. I'm pushing this only for reference, you don't need to worry about this (yet).


denschub (commented):
@ksy36 rightfully mentioned that this may cause an issue, because there already is another bancosantander UA override shipped. @denschub, please make sure this works without issues!

ksy36 (approved):
Looks good to me!

PR #255, opened 2021-12-10
Bug 1745313 - Remove Slack User Agent override.

denschub:
Base is `release-hotfix`, not `main`, because that is the first target. Patchlevel version bump, this is planned to be shipped as a System Addon Update.

r? @wisniewskit


wisniewskit (approved):

JulienRobitaille:
Hi, I am leaving this here for others that are wondering why slack calls and huddle stopped working again yesterday.
https://bugzilla.mozilla.org/show_bug.cgi?id=1727100

I hope that they publish the update soon.

denschub:
@JulienRobitaille are you on Linux by any chance? Huddles should work just fine on Windows and macOS.

denschub:
We did notice several weird edge-cases with the spoofed UA, so the current state is preferred. You can obviously do whatever you want, and spoof your own UA if that is better for you, but it didn't make sense for us to keep it around for all Firefox users. :) Thanks for getting back to me, though!

PureTryOut:
Sorry for bumping such an old thread, but Slack Huddles are still broken on Firefox on Linux, almost 2 years later! I'm sure everyone could just add an user-agent override locally but it should really work out of the box, not everyone has the technical knowledge needed for manual user-agent overrides.

Can this user-agent override be re-added, but for Linux only maybe? Preferably Slack just enables huddles on Linux of course, not sure why they don't, but at least re-instating the previous workaround would improve things already. Sure some weird edge-cases might re-appear but I'd argue that huddles not working is a more severe problem, it's literally the only reason I still have the desktop client installed.

ghomem:
Still broken on Firefox 117 on Ubuntu. What is the simplest way to enable Slack Huddles?

puresick:
This is ultimately an issue of Amazons Chime SDK which is being used by Slack.
I had opened an issue over there some time ago – leaving it here as a reference: https://github.com/aws/amazon-chime-sdk-js/issues/2044

ghomem:
thanks a lof for the information @puresick .  Is there a workaround for Firefox users on Linux?

@denschub

ghomem:
I usually avoid installing extensions but I see that the native way of doing this was removed :( 

https://bugzilla.mozilla.org/show_bug.cgi?id=1589607

If that extension is trustworthy I can give it a try.

puresick:
I cannot guarantee you that, but you could look into the source code if you know your way around HTML, CSS and JavaScript: https://github.com/ray-lothian/UserAgent-Switcher/

In general, Add-Ons with the "Recommendation" tag are, AFAIK, reviewed by the Firefox team to meet their performance and security standards.

ghomem:
I tried it and it works, as in it makes the Huddle widget appear. I used whitelist mode and added only slack.com the whitelist.

But it has a terribly counterintuitive UX, way worse than the simple definition of a per site pref that has been removed by Mozilla (see previous comment). I had to use the logs of my own webserver to make sure that I was configuring it right.

I still don't understand why this has been removed from about:compat since it could be turned on/off easily there. They could even ship it as off by default.

I really want to use Firefox but even Mozilla is making it difficult.

I will report later on whether the actual calls work.

Big thanks you you @puresick .

ghomem:
Tried this on a 1-1- call and:

* sound + remote camera worked
* my sound worked  but my did not (after several permissions where I authorized Firefox to use it)
* I could not share my screen (the button did not do anything)

ghomem:
Thank you for the clarifications @denschub . I apologize for my accumulated frustration on this issue. I already "complained" to Amazon and Slack, of course. The problem is a combination of things. This

https://bugzilla.mozilla.org/show_bug.cgi?id=1589607

combined with the bad usability of the extension. Of course at the root of all this is Slack, but I will not use this space to expand on that.

Thank you again.

denschub:
Here's a suggestion: building a Web Extension isn't that hard for anyone with a bit of basic web development knowledge. There's an [API to change request headers](https://developer.mozilla.org/en-US/docs/Mozilla/Add-ons/WebExtensions/API/webRequest/onBeforeSendHeaders) that allows you to change the User-Agent header for Slack specifically. You might also need to change some of the `navigator` properties, but that can be done with [a content script](https://developer.mozilla.org/en-US/docs/Mozilla/Add-ons/WebExtensions/Content_scripts), and this PR should give a good inspiration (everything in this repo is licensed under the MPL). You can also take a bit of inspiration from the ["Google Search Fixer"](https://addons.mozilla.org/en-US/android/addon/google-search-fixer/) addon for Firefox for Android (source is [here](https://github.com/wisniewskit/google-search-fixer)).

Go ahead and build a Web Extension that spoofs as Chrome on Firefox Linux for Slack only. Upload it to addons.mozilla.org, and link it here. That way, nobody has to deal with bad UX in UA spoof extensions, and instead they can just install an addon if they want. But also, that way, we're not getting the one receiving bug reports if things go wrong.

ghomem:
That's not a bad idea, but having a per-domain pref is a much better idea. That's because the distribution of extensions is very different from the distribution of the base browser. For instance, I acquire Firefox via Canonical which I accept as a trusted supplier. Adding extensions means adding something else to the suppliers list. It is very different from a security perspective. The difference is the same that makes me resistant to install the Slack application directly and pushes me to use the browser instead. 

(This is not the place to develop risk management conversations but I assume you understand what I mean)

From what I could see in:

https://bugzilla.mozilla.org/show_bug.cgi?id=1513574

the option to set a per-domain preference was removed because

Screenshot_20240117_220409

But this seems to be a misunderstanding because the per-domain preference was generic whereas the webcompat addon includes only a whitelist of interventions that are known to work.

So, a generic preference where the user could take the risk to make a change was removed in favor of a restricted whitelist where the user has no control.

The impacts of changing a generic preference on Firefox are something that the user is accountable for just as the impacts of installing a specific extension - bug reports from generated by such actions "it things go wrong" are equally unfair. But the effort and the security risk associated to change a generic pref are much lower.

In short: it would be nice if Firefox brought back the per-domain UA override feature which had better usability and no security risks, as far as I understand.

denschub:
> a generic preference where the user could take the risk to make a change was removed

Absolutely nothing on `about:config` is designed for the user to fiddle with.

`UserAgentOverrides.jsm`, even if it were still here, would not help you in this case. You also need to change `navigator.userAgent`, which [the old impl did not do](https://bugzilla.mozilla.org/show_bug.cgi?id=1084475). You'd also need to change another `navigator` property, as did happen in this PR - again, something that the old system never could do. For a UA override to work in cases like Slack, you need a bit more powerful APIs, which is why we replaced `UserAgentOverrides.jsm` with this very system addon you're commenting in. We also replaced it because the old method was slow and buggy, but I'm sure you have figured it out given you just screenshotted one of the bugs in question. There is no point in maintaining two system features that can do the same from our point of view (which, again, is what all of those methods are designed for - they're explicilty not designed for users to fiddle with).

I also explained how you can do that yourself, without having to trust anyone except yourself - because all the relevant APIs are available to anyone. If you don't want to do that, that's fine - but that's not my fault, and I'm not going to suggest to my team that we have to maintain known-problematic UA overrides - and I'm also not going to suggest to the Networking module engineers that we have to add a second mechanism to change UA strings. Firefox engineers don't need another mechanism, and you don't need one either, as that's what we have an entire team maintaining extension APIs for. We put those things into the browser so people like you can use them if you want to do more/different things than we do.

Everyone is invited to use the existing WebExtension APIs to mimic what we're doing here. You might make Slack work fine *for you*, you might break it horribly, but either way, you did it to yourself. This discussion is on a PR from 2021. The PR has been merged, and there won't be a revert of this removal. This is the end of the discussion in this PR.

PR #265, opened 2022-02-17
Issue #256 - Add tooling to ensure and assist with the package.json version number

denschub:
To us, the version number in `package.json` is pretty useless. We used to keep this manually in sync with the `manifest.json`s setting, but in https://github.com/mozilla-extensions/webcompat-addon/commit/7a445d242a8354c1b9347a60d415932091f753dc, I decided to stop that and just have it say `1.0.0` all the time, because to us, the relevant information is in `manifest.json` only.

Unfortunately, this causes an issue. CI uses the `package.json`s version number to set the version number in .xpi, and thus this is also relevant during rollouts. I discovered this during the recent Release-Hotfix, where the rollout systems wanted to deploy a version `1.0.0`. I talked to the Addons Pipeline team, and while they agreed that this is not perfect, reading the version from `manifest.json` isn't really an option either, since some System Addons contain multiple manifest files that get adjusted during build steps, and there generally are a lot of edge-cases.

So the best solution is for us to keep `manifest.json` and `package.json` always in sync. Since a mismatch can make a Release Hotfix more complicated, which usually isn't the time where you want to debug and manually adjust things due to the general pressure related to release hotfixes, this PR adds

- an extra step to the `lint` task that verifies that the version numbers match up, and if not, fails hard and thus will turn CI red.
- an extra step of the auto formatting pipeline in addition to running `prettier`, which automatically overwrites the version number in `package.json` with whatever is defined in `manifest.json`, so we don't have to manually do this.

---

Notice to @karlcow, @ksy36, and @wisniewskit: **This PR might break your workflow**:

- If you have set up the `pre-push` hook [as described in `README.md`](https://github.com/mozilla-extensions/webcompat-addon#run-tests-and-codestyle-checks-automatically-before-pushing), you will no longer be able to push if the version numbers do not match up. Instead, the push will fail, showing a `!! Running codestyle checks returned an error! Please run 'npm run lint' to see what's wrong.` error message. If you have not yet set up the `pre-push` hook, please do so: it's the simplest way to ensure that CI doesn't unexpectedly fail with linting errors, and it runs pretty fast.
- If you're currently using `npm run prettier` to have prettier auto-format all files, this task no longer exists, and you will see a `Missing script: "prettier"` error. Instead, please run `npm run autoformat` from now on, which will run `prettier`, but it will also auto-set the version number in `package.json`, so you don't have to do that manually. 

---

This closes #256

PR #266, opened 2022-02-17
No bug - Update taskgraph revision.

denschub:
This is blocking #265, see https://github.com/mozilla-extensions/xpi-manifest/pull/152


denschub:
Merging with r=me, this is a CI-only change, and CI is still green.

PR #280, opened 2022-05-25
Bug 1752989 - Ship v102.0.0 of the WebCompat System Addon

ksy36:
Overrides and interventions for 102 release https://bugzilla.mozilla.org/show_bug.cgi?id=1752989


src/data/injections.js
```diff
@@ -123,15 +123,15 @@ const AVAILABLE_INJECTIONS = [
   },
   {
     id: "bug1561371",
-    platform: "android",
-    domain: "mail.google.com",
+    platform: "desktop",
+    domain: "cloud.google.com",
     bug: "1561371",
```
wisniewskit, line 128: We might want to update this to whatever the right bug # is for the cloud issue, rather than leaving it pointing to the gmail issue.

src/manifest.json
```diff
@@ -2,7 +2,7 @@
   "manifest_version": 2,
   "name": "@EXTENSION_NAME@",
   "description": "Urgent post-release fixes for web compatibility.",
-  "version": "101.7.0",
```
wisniewskit, line 5: This should change to 101.8.0 as I'm landing a patch against SmartBlock right now. Or, you could backport that patch as part of this set, if you'd rather do that (https://phabricator.services.mozilla.com/D147309)

wisniewskit (approved):

PR #283, opened 2022-06-23
Bug 1752990 - Interventions for Firefox 103!

denschub:
The usual fun! This set includes three patches from @wisniewskit, which he published at:

- https://phabricator.services.mozilla.com/D150040
- https://phabricator.services.mozilla.com/D150065
- https://phabricator.services.mozilla.com/D150063

I'm including these here, and I'll also land them, to reduce the potential for head-on collisions. :)

r? @ksy36, and also r? @wisniewskit just to make sure I didn't break your changes (I only could test Imgur, that works, I didn't dig into the other two patches!)


src/data/injections.js
```diff
@@ -496,10 +496,15 @@ const AVAILABLE_INJECTIONS = [
...
-    domain: "draft.js",
+    domain: "Sites using draft.js",
     bug: "1739489",
     contentScripts: {
-      matches: ["*://draftjs.org/*", "*://www.facebook.com/*"],
+      matches: [
+        "*://draftjs.org/*", // Bug 1739489
+        "*://www.facebook.com/*", // Bug 1739489
+        "*://twitter.com/*", // Bug 1776229
+        "*://mobile.twitter.com/*", // Bug 1776229
```
denschub, line 506: Note: I'm explicitly adding `mobile.` as its own entry instead of just overriding `*.`, just to be on the safe side. This is Twitter, after all... :)

ksy36 (approved):
LGTM:) Thanks!

PR #285, opened 2022-07-08
Fixes #249 - Update the default Chrome version for UA overrides in getDeviceAppropriateChromeUA

### LINKED ISSUE #249: Investigate updating the default Chrome version for UA overrides
From a Chat with Tom on 2021-09-27:

> btw, i just noticed we're still mimicking chrome 76.0.3809.111 in getDeviceAppropriateChromeUA()... i'm surprised that hasn't needed an update so far

and he's not wrong, that version is quite old. In https://github.com/mozilla-extensions/webcompat-addon/commit/44ce3f4507f2c25b7817697db31afce246ff7621, I added the possibility to specify the used version number for each override, because I wanted to spoof as Chrome 95 for Slack.

We should be updating the Chrome version eventually, mainly to avoid exposing our users to the potential of "your browser is too old" warnings. We should address that before it becomes an issue. Chrome 76 was released in July 2019, so that's a real possibility now.

Unfortunately, blindly flipping the version isn't a good idea, we should probably re-test all interventions to make sure that a) they still work, and b) the site doesn't break in other ways. If we don't want to spend lots of time testing, this is probably blocked by https://github.com/mozilla/webcompat-team-okrs/issues/203 for now, but I wanted to make sure we have at least an issue for that.

ksy36:
I've retested all overrides that are using `getDeviceAppropriateChromeUA` and all work as expected with newer version, here is the list:

site | device
-- | --
wp1-ext.usps.gov | - android
autotrader.ca | - android
lffl.org | - android
app.xiaomi.com | - android
www.dealnews.com | - android
apppmedia.jp | - android
covid.cdc.gov | - both
vmos.cn | - android
yebocasino.co.za | - android
automesseweb.jp | - android
slrclub.com | - android
workflow.base.vn | - android
nordjyske.dk | - android
www.dolcegabbana.com | - android
www.otsuka.co.jp | - desktop
animalplanet.com | - android
frontgate.com | - android

It would be nice to update version for interventions that don't use `getDeviceAppropriateChromeUA` function and have hardcoded UA strings. I've filed a separate issue for that https://github.com/mozilla-extensions/webcompat-addon/issues/286

### DIFF
--- a/src/data/ua_overrides.js
+++ b/src/data/ua_overrides.js
@@ -838,7 +838,10 @@ const AVAILABLE_UA_OVERRIDES = [
     config: {
       matches: ["*://nordjyske.dk/*"],
       uaTransformer: originalUA => {
-        return UAHelpers.getDeviceAppropriateChromeUA("97.0.4692.9", "Pixel 4");
+        return UAHelpers.getDeviceAppropriateChromeUA(
+          "103.0.5060.71",
+          "Pixel 4"
+        );
       },
     },
   },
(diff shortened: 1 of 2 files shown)


wisniewskit (approved):
LGTM. Thanks a bunch!

PR #287, opened 2022-07-21
V104.0.0 interventions, injections and shims

wisniewskit:
Bugzilla meta-bug: https://bugzilla.mozilla.org/show_bug.cgi?id=1776673

(Note that the Firebase shim update is a backport, and has already landed on mozilla-central)


wisniewskit:
Re-making this to split up a patch.

PR #288, opened 2022-07-21
V104.0.0 interventions, injections and shims

wisniewskit:
Bugzilla meta-bug: https://bugzilla.mozilla.org/show_bug.cgi?id=1776673

(Note that the Firebase shim update is a backport, and has already landed on mozilla-central)

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,12 +1,12 @@
 {
   "name": "webcompat",
-  "version": "103.13.0",
+  "version": "104.0.0",
   "lockfileVersion": 2,
   "requires": true,
   "packages": {
     "": {
       "name": "webcompat",
-      "version": "103.13.0",
+      "version": "104.0.0",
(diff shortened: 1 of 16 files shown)


denschub (changes requested):
There are two Vim Swapfiles again! 

```bash
cat '*.swp' >> ~/.gitignore-global
git config --global core.excludesfile ~/.gitignore-global
```

and you'll never see this comment again. ;D

Besides that, looks good. The YouTube intervention has the potential of triggering perf test because it adds a network listener, but I hope this isn't the case.

denschub (approved):

denschub:
I'm not entirely sure why the `dep-signing-webcompat` task fails, but looking at the logs, it seems unrelated to your changes. Feel free to merge and land, and I'll figure out why it's broken if it stays that way.

PR #291, opened 2022-08-16
Ship v105.0.0 of the WebCompat System Addon

ksy36:
r? @wisniewskit


wisniewskit (approved):
r+, but we ought to check if the seznam fix is good on desktop too (for narrow viewports, as it might be a responsive design)

PR #294, opened 2022-09-08
Bug 1776675 - Interventions for v106

denschub:
Surprise, it's one week early! (Okay, not a surprise, since I'll be on PTO next week, but hey..)

r? @wisniewskit


wisniewskit (approved):
LGTM!

PR #297, opened 2022-10-31
Adjust export tooling for the new firefox-android monorepo.

denschub:
This adjusts the export tooling to be able to export into the new [firefox-android monorepo](https://github.com/mozilla-mobile/firefox-android), since the old android-components repo is now archived.

## 🚨 This will break your workflow, and maybe your heart if you don't pay attention!

While the jake export job is still called `export-ac`, I renamed the environment variable used to indicate the path of the repo. What used to be called `EXPORT_AC_LOCATION` is now called `EXPORT_ANDROID_MONOREPO_LOCATION`, and it should point to the **root directory** of your local `firefox-android` clone!

If you have set this variable in your `.zshrc` or `.bashrc` or whereever, **now** is a good chance to adjust that!

Explicitly pinging both @ksy36 and @wisniewskit for visibility. Note that the old workflow will no longer work, so you have to adjust your ENV vars before you touch the addon the next time!

r? whoever reviews first.

### DIFF
--- a/Jakefile
+++ b/Jakefile
@@ -87,16 +87,18 @@ function getMozillaCentralLocation() {
  * @returns {string} the location of AndroidComponents
  */
 function getAndroidComponentsLocation() {
-  let acLocation = path.resolve(
-    process.env.EXPORT_AC_LOCATION || "../android-components"
+  let monorepoLocation = path.resolve(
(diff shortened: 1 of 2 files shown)


ksy36 (approved):
thanks for looking into this :)

denschub:
I've merged this now. Since it's Tom's release cycle next, he will notice this change either way. ;D

PR #304, opened 2023-03-01
Patches for for v112.0.0

ksy36:
Patches for https://bugzilla.mozilla.org/show_bug.cgi?id=1805409


src/data/injections.js
```diff
@@ -801,14 +761,106 @@ const AVAILABLE_INJECTIONS = [
   },
   {
     id: "bug1811325",
-    platform: "all",
+    platform: "desktop",
     domain: "www.bdo.com.ph",
```
ksy36, line 765: I've changed this to be desktop only as Calin mentioned both sites work fine on Android without the intervention

src/injections/js/bug1818818-fastclick-legacy-shim.js
```diff
@@ -0,0 +1,20 @@
...
+/* globals exportFunction */
+
+const proto = CSS2Properties.prototype.wrappedJSObject;
+Object.defineProperty(proto, "msTouchAction", {
+  get: exportFunction(function() {
+    return "none";
+  }, window),
+
+  set: exportFunction(function() {}, window),
+});
```
ksy36, line 20: This is simpler variation of a fastclick patch for an older version of fastclick that don't use `touchAction` but only `msTouchAction` (https://bugzilla.mozilla.org/show_bug.cgi?id=1818818#c0)

wisniewskit (approved):

PR #313, opened 2023-06-01
Patches for version 115.0.0

wisniewskit:


src/data/ua_overrides.js
```diff
@@ -1264,6 +1264,23 @@ const AVAILABLE_UA_OVERRIDES = [
...
+     * The site's content is not loaded without a Chrome UA spoof.
+     */
+    id: "bug1836116",
+    platform: "all",
+    domain: "www.slushy.com",
+    bug: "1836116",
+    config: {
+      matches: ["*://www.slushy.com/*"],
+      uaTransformer: originalUA => {
+        return originalUA + " Chrome/15.0.0.0";
```
denschub, line 1280: Is version 15 a typo or deliberate? Asking only because this is a version from 2011. :)

src/data/ua_overrides.js
```diff
@@ -1281,6 +1281,23 @@ const AVAILABLE_UA_OVERRIDES = [
...
   },
+  {
+    /*
+     * Bug 1836135 - UA override for gts-pro.sdimedia.com
+     *
+     * The site's content is not loaded without a Chrome UA spoof.
+     */
+    id: "bug1836135",
+    platform: "all",
+    domain: "www.slushy.com",
```
denschub, line 1292: Wrong domain here, this should be `gts-pro.sdimedia.com` or maybe `sdimedia.com`. Not that this has any real impact besides our listing, but.

src/injections/js/bug1836157-thai-masszazs-niceScroll-disable.js
```diff
@@ -0,0 +1,17 @@
+/**
```
denschub, line 1: Please add a license header like

https://github.com/mozilla-extensions/webcompat-addon/blob/cdc9edd6858a9713a39d0e441afdce1a1647f721/src/injections/js/bug0000000-testbed-js-injection.js#L1-L3

to this intervention and all others as well (including CSS).

denschub (changes requested):
Some nits!

src/data/ua_overrides.js
wisniewskit: It's just the first version I tried. And it confirms that we don't have to worry about keeping the version number up-to-date, so I left it at that. I can try with 0.0.0.0, or even use the frozen version number Chrome has decided to use instead?

denschub: Chrome's frozen number would probably be the best solution (and it would mean we don't have to worry).

src/data/ua_overrides.js
wisniewskit: Good eye, will fix.

denschub (approved):
🚢

PR #316, opened 2023-06-29
Bug 1836521 - Intervention for v116

denschub:
My brain is fried, and there's a bunch of holiday/vacation stuff coming up until the merge day, so let's just ship this minimal release cycle. I'll probably slot some other things into the GitHub repo next week, but I don't want to delay shipping this in the right train.


wisniewskit (approved):
Leeeet's do this

PR #317, opened 2023-07-06
Bug 1841991 - Ship flexbox paint order intervention for wilton.com.

denschub:
I'm feeling adventurous today, so I decided to build the intervention right after diagnosis. I'm not sure if this is a good idea because it reduces context switches (and the need for cross-people information transfer), or if this is a bad idea because it fragments review time and may create merge conflicts.

Either way, r? @wisniewskit (it's your cycle).

### DIFF
--- a/src/data/injections.js
+++ b/src/data/injections.js
@@ -1012,6 +1012,20 @@ const AVAILABLE_INJECTIONS = [
       ],
     },
   },
+  {
+    id: "bug1841991",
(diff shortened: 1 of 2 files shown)


wisniewskit (approved):
r+, thanks!

wisniewskit:
Since there is a merge conflict, I'll just add this to the roll-up patch for 117 manually.

PR #322, opened 2023-09-06
apply ua override for gbf.game.mbga.jp

rhjdvsgsgks:
it is another domain of gbf. mabe this compat workaround also works for it. but i haven't tested yet. because i have only a phone at this moment

see also: https://github.com/webcompat/web-bugs/issues/34310#issuecomment-1708967412


denschub (approved):
Thanks for the PR, @rhjdvsgsgks! :)

I'll merge this right now so that it doesn't get missed in the next cycle. FYI, @wisniewskit.

PR #328, opened 2023-10-31
Bug 1862225 - Change UA override for tesco.com

ksy36:


src/data/ua_overrides.js
```diff
@@ -1379,21 +1379,20 @@ const AVAILABLE_UA_OVERRIDES = [
...
-      matches: [
-        "*://*.tesco.com/*", // #1858664
-      ],
+      matches: ["*://*.tesco.com/*"],
       uaTransformer: originalUA => {
-        return UAHelpers.capRvTo109(originalUA);
+        const cappedRv = UAHelpers.capRvTo109(originalUA);
+        return UAHelpers.capVersionToNumber(cappedRv);
       },
     },
```
ksy36, line 1397: The issue is reproducible on all platforms and not just Windows as I initially thought :|
So capping the rv to 109.0 and version to 120.0 for all platforms

wisniewskit (approved):
Thanks!

ksy36:
thank you:)

PR #335, opened 2024-03-12
Bug 1876363 - Ship v125.0.0 of the WebCompat System Addon

ksy36:
https://bugzilla.mozilla.org/show_bug.cgi?id=1876363


src/data/injections.js
```diff
@@ -540,19 +508,19 @@ const AVAILABLE_INJECTIONS = [
         "*://*.avanabayview.com/*", // #118617
         "*://*.breakpointeandcoronado.com/*", // #117735
         "*://*.courtsatspringmill.com/*", // #128404
+        "*://*.fieldstoneamherst.com/*", // #131703
```
wisniewskit, line 511: I think this is supposed to be 132974, yes?

wisniewskit (approved):
Just one minor note about a mismatched bug number, otherwise good to go!

src/data/injections.js
ksy36: good catch, thanks!

PR #336, opened 2024-03-21
No bug - Adjust export tooling for the in-tree Android Components.

denschub:
This also adds a new command, `npm run jake export`, that will export both mobile and desktop. Since those two are in the same repository anyway, I don't see a good reason to separate them in most use-cases. The individual commands still exist, though, in case you only need to update one.

r? @ksy36 
fyi! @wisniewskit

### DIFF
--- a/Jakefile
+++ b/Jakefile
@@ -83,28 +83,6 @@ function getMozillaCentralLocation() {
 }
 
 /**
- * @throws if Android-Components could not be found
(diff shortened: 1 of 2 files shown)


wisniewskit:
Thank you for this!

wisniewskit (approved):

ksy36:
Thank you, LGTM!

PR #342, opened 2024-05-24
Bug 1860417 - Add intervention for samsung.com on Android

cpeterso:
Apply bug 1799968's intervention for samsung.com on Linux to Android.


denschub (approved):
An alternative approach could have been to ship this to `all` platforms, and then do the "do I need to do something?" check inside the content script. But this is fine, too, it doesn't really matter. :)

r+, we'll get this shipped soon. Thanks for the PR!

PR #355, opened 2024-07-16
Bug 1842767 - Add UA override for passport.bilibili.com

wisniewskit:


src/data/ua_overrides.js
```diff
@@ -1210,6 +1210,26 @@ const AVAILABLE_UA_OVERRIDES = [
...
+     *
+     * Spoofing as Chrome makes the login page use a mobile layout.
+     */
+    id: "bug1842767",
+    platform: "android",
+    domain: "passport.bilibili.com",
+    bug: "1842767",
+    config: {
+      matches: [
+        "*://*.passport.bilibili.com/",
```
denschub, line 1225: Nit: `*` [matches zero or more characters](https://developer.mozilla.org/en-US/docs/Mozilla/Add-ons/WebExtensions/manifest.json/content_scripts#globs), so we don't actually need this.

denschub (approved):
One nit I'd like to have fixed, but r+ otherwise.

PR #358, opened 2024-07-18
Unlock scrolling on gracobaby.ca and softrans.ro

wisniewskit:


src/data/injections.js
```diff
@@ -1111,6 +1111,34 @@ const AVAILABLE_INJECTIONS = [
       ],
     },
   },
+  {
+    id: "bug1896571",
+    platform: "all",
```
ksy36, line 1116: wonder if this should be android only?

ksy36 (approved):

PR #362, opened 2024-07-23
Fixes for 129.9.0 (digits.t-mobile.com, unimedlitoral.com.br, lg.jio.com)

wisniewskit:


denschub (approved):
Oh, our first client-hints adjacent intervention. How exciting! 🙃

src/data/ua_overrides.js
```diff
@@ -1261,6 +1261,42 @@ const AVAILABLE_UA_OVERRIDES = [
...
+     * Site incorrectly blocks Firefox on Android. A desktop UA works.
+     */
+    id: "bug1902474",
+    platform: "android",
+    domain: "lg.jio.com",
+    bug: "1902474",
+    config: {
+      matches: ["*://lg.jio.com/*"],
+      uaTransformer: () => {
+        return navigator.userAgent.replace(/ (Mobile|Tablet);/, "");
```
denschub, line 1296: Uh, wait, this doesn't actually work. The original UA is passed into this function as an argument, and you have to return the new one.

```
      uaTransformer: (originalUA) => {
        return originalUA.replace(/ (Mobile|Tablet);/, "");
      }
```

denschub (changes requested):

src/data/ua_overrides.js
denschub: Oh, true, the script has access to it. But the tests fail, as the `navigator` object on the global isn't set there, so we should still update.

denschub (approved):

PR #379, opened 2024-10-29
Bug 1927603 - v134.0.0 updates for webcompat interventions

wisniewskit:


denschub (approved):
Looks all good to me!

PR #383, opened 2024-11-13
V134.4.0

wisniewskit:


denschub (changes requested):
This PR does not match the stack ending with D228929. The MPL header in `injections/js/bug1930440-online.singaporepools.com-prevent-unsupported-alert.js` is missing from this PR.

denschub (approved):

